PR #7, opened 2020-04-27
README: better explain the known bugs

doronbehar:
Also, use --version and not only [version] in usage.

I missed this info as a first time user.


Mic92:
Thanks!

PR #25, opened 2020-11-18
Add test flag

### LINKED ISSUE #24: Tests
Just found out about this and now I am intrigued :D

Would it be possible to add running `passthru.tests` for a derivation being updated before committing?

Example:

`nix-update --commit --test sbt` and the commit step will only run if the tests have passed.

NeQuissimus:
Closes #24 

This is sort of what I had in mind to add testing.
Might need some `is_nix_flakes()` love, I am not super familiar with that.

### DIFF
--- a/nix_update/__init__.py
+++ b/nix_update/__init__.py
@@ -7,7 +7,7 @@ from typing import NoReturn, Optional
 from .eval import Package
 from .options import Options
 from .update import update
-from .utils import run, current_system
+from .utils import run
 
 
 def die(msg: str) -> NoReturn:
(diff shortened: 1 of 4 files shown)


nix_update/__init__.py
```diff
@@ -132,14 +136,29 @@ def nix_build(options: Options) -> None:
...
 
+def nix_test(git_dir: str, package: Package, system: Optional[str]) -> None:
+    if not package.tests or not system:
+        die(f"Package {package.name} does not define any tests for {system}")
+    else:
+        tests = []
+        for t in package.tests:
+            tests.append("-A")
+            tests.append(f"tests.{t}.{system}")
+        cmd = ["nix-build", f"{git_dir}/nixos/release.nix"] + tests
```
Mic92, line 147: Not all tests are nixos tests. Sometimes people just reference derivations here. I will have a look at that.

Mic92:
I switched now to nixFlakes by default to keep the test matrix small.

nix_update/__init__.py
NeQuissimus: Interesting, I don't think I have ever seen that. I basically tried to do something similar to @ofborg here

NeQuissimus:
Rebased on top of `master`

nix_update/__init__.py
Mic92: It's an edge case I have seen it only a handful times.

Mic92:
It's now a bit shorter and works with other package set that are not nixpkgs.

nix_update/__init__.py
```diff
@@ -139,29 +139,26 @@ def nix_build(options: Options) -> None:
...
-        cmd = ["nix-build", f"{git_dir}/nixos/release.nix"] + tests
-        run(cmd)
+def nix_test(package: Package) -> None:
+    if not package.tests:
+        die(f"Package '{package.name}' does not define any tests")
+
+    tests = []
+    for t in package.tests:
+        tests.append("-A")
+        tests.append(f"{package.attribute}.tests.{t}")
```
NeQuissimus, line 149: Ah yeah, that's clever. Much better :)

NeQuissimus:
@Mic92 Any chance you could tag this, so we can update the version in nixpkgs? :)

Mic92:
https://github.com/Mic92/nix-update/releases/tag/0.2

Mic92:
Remember to also use nixFlakes in nixpkgs when updating. Older nix versions no longer work.

PR #27, opened 2020-11-27
Skip updating hashes without version change

NeQuissimus:
This is going to require an opinion, I think :)

I don't know if you have a use case for updates that only change the hash (seems weird to me but maybe there is a reason).
Anyways, I have a use case against that:

Maintainers on different nix versions currently swap between the old SHA256 and the new notation.
This confuses `nix-update` and triggers commits that purely change the hash format (but get committed as `PACKAGE: 1.2.3 -> 1.2.3` with `--commit`).

This change will not trigger hash changes if the version has not been updated.

### DIFF
--- a/nix_update/update.py
+++ b/nix_update/update.py
@@ -11,7 +11,7 @@ from .version import fetch_latest_version
 from .git import old_version_from_git
 
 
-def replace_version(package: Package) -> None:
+def replace_version(package: Package) -> bool:
     old_version = package.old_version
     new_version = package.new_version
     assert new_version is not None
@@ -25,6 +25,8 @@ def replace_version(package: Package) -> None:
                 print(line.replace(old_version, new_version), end="")
     else:
         info(f"Not updating version, already {old_version}")
+
+    return old_version != new_version
 
 
 def to_sri(hashstr: str) -> str:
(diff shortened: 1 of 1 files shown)


Mic92:
Actually I use this sometimes and than fix the version number manually :). Sometimes I have updated the version number upfront. But I see your point.

NeQuissimus:
I think the way I set the default boolean, the functionality should still be available if `opts.version` is `skip`

Mic92:
@NeQuissimus how do you like this solution? https://github.com/Mic92/nix-update/pull/28

NeQuissimus:
@Mic92 It won't help with people not running `nix-updates` but updating manually. But I think I can live with that. One additional thought: Is there a way we could teach the git commit logic about this?
Instead of `pkg: 1.2.3 -> 1.2.3`, could you make it say `pkg: Normalize hash` or something like that?

Mic92:
I will scrape the version diff from: `git diff --word-diff=porcelain`

Mic92:
> @Mic92 It won't help with people not running `nix-updates` but updating manually. But I think I can live with that. One additional thought: Is there a way we could teach the git commit logic about this?
> Instead of `pkg: 1.2.3 -> 1.2.3`, could you make it say `pkg: Normalize hash` or something like that?

I think after https://github.com/Mic92/nix-update/pull/30 we add the logic you proposed.

PR #28, opened 2020-11-29
normalize hashes

Mic92:


Mic92:
Another important aspect of this update is that it makes normalizes hashes across different nix versions.

PR #30, opened 2020-12-01
get old version from git

Mic92:


Mic92:
my local black version is different from CI.

PR #37, opened 2021-01-18
Add --unstable-version argument

NeQuissimus:
Enabling this flag will include all versions
containing '-' and '+' from being considered valid.
As per SemVer, these are the separators for unstable
versions (i.e. "-RC2", "+abcdef" suffixes).

### DIFF
--- a/nix_update/__init__.py
+++ b/nix_update/__init__.py
@@ -33,6 +33,11 @@ def parse_args() -> Options:
         default="(.*)",
     )
     parser.add_argument(
+        "--unstable-version",
+        action="store_true",
+        help="Include semantic versions with build or pre-release suffixes",
+    )
+    parser.add_argument(
         "--run",
         action="store_true",
         help="provide a shell based on `nix run` with the package in $PATH",
@@ -55,6 +60,7 @@ def parse_args() -> Options:
         attribute=args.attribute,
         test=args.test,
         version_regex=args.version_regex,
+        unstable_version=args.unstable_version,
     )
 
 
--- a/nix_update/__init__.py
+++ b/nix_update/__init__.py
@@ -33,11 +33,6 @@ def parse_args() -> Options:
         default="(.*)",
     )
     parser.add_argument(
-        "--unstable-version",
-        action="store_true",
-        help="Include semantic versions with build or pre-release suffixes",
-    )
-    parser.add_argument(
         "--run",
         action="store_true",
         help="provide a shell based on `nix run` with the package in $PATH",
@@ -46,7 +41,7 @@ def parse_args() -> Options:
         "--shell", action="store_true", help="provide a shell with the package"
     )
     parser.add_argument(
-        "--version", nargs="?", help="Version to update to", default="auto"
+        "--version", nargs="?", help="Version to update to", default="stable"
     )
     parser.add_argument("attribute", help="Attribute name within the file evaluated")
     args = parser.parse_args()
@@ -60,7 +55,6 @@ def parse_args() -> Options:
         attribute=args.attribute,
         test=args.test,
         version_regex=args.version_regex,
-        unstable_version=args.unstable_version,
     )
 
 
--- a/nix_update/options.py
+++ b/nix_update/options.py
@@ -12,3 +12,4 @@ class Options:
     run: bool = False
     build: bool = False
     test: bool = False
+    unstable_version: bool = False
--- a/nix_update/options.py
+++ b/nix_update/options.py
@@ -12,4 +12,3 @@ class Options:
     run: bool = False
     build: bool = False
     test: bool = False
-    unstable_version: bool = False
--- a/nix_update/options.py
+++ b/nix_update/options.py
@@ -4,7 +4,7 @@ from dataclasses import dataclass
 @dataclass
 class Options:
     attribute: str
-    version: str = "auto"
+    version: str = "stable"
     version_regex: str = "(.*)"
     import_path: str = "./."
     commit: bool = False
--- a/nix_update/utils.py
+++ b/nix_update/utils.py
@@ -1,5 +1,4 @@
 import os
-import re
 import subprocess
 import sys
 from pathlib import Path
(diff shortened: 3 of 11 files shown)


nix_update/version/gitlab.py
```diff
@@ -29,7 +31,9 @@ def fetch_gitlab_version(url: ParseResult, version_regex: str) -> Optional[str]:
             name = tag["name"]
             assert isinstance(name, str)
             extracted = extract_version(name, version_regex)
-            if extracted is not None:
+            if extracted is not None and (
+                (not version_stable) or version_is_stable(extracted)
```
Mic92, line 35: I remember that the github also had a field if the version is stable.

nix_update/utils.py
```diff
@@ -46,3 +46,7 @@ def extract_version(version: str, version_regex: str) -> Optional[str]:
         if group is not None:
             return group
     return None
+
+
+def version_is_stable(version: str) -> bool:
+    return version is not None and "-" not in version and "+" not in version
```
Mic92, line 52: I am not so sure if this is a good heuritic. I would rather look for `rc`, `alpha`, `beta`, `nightly`.

nix_update/__init__.py
```diff
@@ -32,6 +32,11 @@ def parse_args() -> Options:
         help="Regex to extract version with, i.e. 'jq-(.*)'",
         default="(.*)",
     )
+    parser.add_argument(
+        "--stable-version",
+        action="store_true",
+        help="Only use semantic versions without build or pre-release suffixes",
```
Mic92, line 38: I wonder if the behavior should be inverted and the flag could be called `unstable` or `prerelease`.

nix_update/version/github.py
```diff
@@ -33,7 +35,11 @@ def fetch_github_version(url: ParseResult, version_regex: str) -> Optional[str]:
     releases = tree.findall(".//{http://www.w3.org/2005/Atom}entry")
     entries = [version_from_entry(x) for x in releases]
     filtered = [
-        x for x in entries if x is not None and extract_version(x, version_regex)
+        x
+        for x in entries
+        if x is not None
+        and extract_version(x, version_regex)
+        and ((not stable_version) or version_is_stable(x))
```
Mic92, line 42: I would actually keep the latest unstable version here. If no stable versions is found an error message could be printed that an unstable version could be used if the flag is given.

Mic92:
> Enabling this flag will exclude all versions
> containing '-' and '+' from being considered valid.
> As per SemVer, these are the separators for unstable
> versions (i.e. "-RC2", "+abcdef" suffixes).

Counter example: Some projects put the project name into the tag i.e. `jq-${version}` or just call it 1.2rc1

nix_update/utils.py
NeQuissimus: Can we not rely on semantic versioning then?

https://semver.org/#spec-item-9 and https://semver.org/#spec-item-10 are what I used for this.

NeQuissimus:
Cases like jq are covered by extracting the version first. `1.2rc1` is just wrong :D But I see the point...

nix_update/__init__.py
NeQuissimus: Oh, we could do that and later expand or introduce a related flag to fetch the HEAD commit from a branch!

nix_update/version/gitlab.py
NeQuissimus: In the API, yes. Not in the Atom feed.

nix_update/utils.py
Mic92: I just grepped for rc1 and the majority of examples did not followed the semver spec. Most of them skipped the `-`.

NeQuissimus: What if I do both? Leave this and also check for the keywords you listed?

Mic92: I am concerned that `-` is often use for something else. I have not seen `+` being used by anything.

Mic92: Ok. `+` is actually used by some Linux distros for example Debian for their internal versioning. I don't think we should use that.

Mic92: But you can also add `preview` to the list.

NeQuissimus: I also added `m[0-9]` for milestones

nix_update/__init__.py
Mic92: Right.

nix_update/__init__.py
```diff
@@ -32,6 +32,11 @@ def parse_args() -> Options:
         help="Regex to extract version with, i.e. 'jq-(.*)'",
         default="(.*)",
     )
+    parser.add_argument(
+        "--unstable-version",
```
Mic92, line 36: Do you think `--version=unstable` could be used for that? Later `--version` could be re-used again to fetch the latest commit. I can also do the refactoring for that.

NeQuissimus: hmmm, possibly... You mean as an alternative to the current `auto`? Might have to rename `auto` to `stable` then.
Maybe that should be a second PR. Not sure...

Mic92: > hmmm, possibly... You mean as an alternative to the current `auto`? Might have to rename `auto` to `stable` then.
> Maybe that should be a second PR. Not sure...

Right. I have a look.

Mic92:
I refactored it a bit since too much duplicated code ended up being in those version fetchers.

nix_update/__init__.py
```diff
@@ -41,7 +41,7 @@ def parse_args() -> Options:
         "--shell", action="store_true", help="provide a shell with the package"
     )
     parser.add_argument(
-        "--version", nargs="?", help="Version to update to", default="auto"
+        "--version", nargs="?", help="Version to update to", default="stable"
```
Mic92, line 44: Mind having a look at the new code? All that it is left to do is to add this to the README.

nix_update/options.py
```diff
@@ -4,11 +4,12 @@
...
-    version: str = "auto"
+    version: str = "stable"
     version_regex: str = "(.*)"
     import_path: str = "./."
     commit: bool = False
     shell: bool = False
     run: bool = False
     build: bool = False
     test: bool = False
+    unstable_version: bool = False
```
Mic92, line 15:
```suggestion
```

NeQuissimus:
LGTM, thanks! I'll make the README additions tomorrow.

PR #41, opened 2021-02-21
Fix git_dir when running without `--commit` argument

NeQuissimus:
The variable is otherwise unset and will cause issues

### DIFF
--- a/nix_update/__init__.py
+++ b/nix_update/__init__.py
@@ -178,8 +178,7 @@ def main() -> None:
     if not os.path.exists(options.import_path):
         die(f"path {options.import_path} does not exists")
 
-    if options.commit:
-        git_dir = validate_git_dir(options.import_path)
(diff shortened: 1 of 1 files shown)


Mic92:
I had this commit flying around... wanted to push it earlier: https://github.com/Mic92/nix-update/pull/43

nix_update/__init__.py
```diff
@@ -178,8 +178,7 @@ def main() -> None:
     if not os.path.exists(options.import_path):
         die(f"path {options.import_path} does not exists")
 
-    if options.commit:
-        git_dir = validate_git_dir(options.import_path)
+    git_dir = validate_git_dir(options.import_path)
```
Mic92, line 181: This basically assumes that there everyone uses git, which might be not the case.

Mic92:
merged in https://github.com/Mic92/nix-update/pull/43 instead.

PR #44, opened 2021-02-21
Fix --format

NeQuissimus:
If used in combination with --commit, the changes
were previously not staged and hence left uncommitted

### DIFF
--- a/nix_update/__init__.py
+++ b/nix_update/__init__.py
@@ -175,7 +175,7 @@ def nixpkgs_review() -> None:
     run(cmd, check=True)
 
 
-def nixpkgs_fmt(git_dir: Optional[str], package: Package) -> None:
+def nixpkgs_fmt(package: Package, git_dir: Optional[str]) -> None:
     cmd = ["nixpkgs-fmt", package.filename]
     run(cmd, check=True)
     if git_dir is not None:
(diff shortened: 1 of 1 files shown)


nix_update/__init__.py
```diff
@@ -216,7 +218,7 @@ def main() -> None:
         nixpkgs_review()
 
     if options.format:
-        nixpkgs_fmt(package)
+        nixpkgs_fmt(git_dir, package)
```
Mic92, line 221: Instead of duplicating code I would just move it before git_has_diff.

nix_update/__init__.py
```diff
@@ -175,9 +175,11 @@ def nixpkgs_review() -> None:
     run(cmd, check=True)
 
 
-def nixpkgs_fmt(package: Package) -> None:
+def nixpkgs_fmt(git_dir: Optional[str], package: Package) -> None:
```
Mic92, line 178:
```suggestion
def nixpkgs_fmt(package: Package, git_dir: Optional[str]) -> None:
```

nix_update/__init__.py
Mic92, line 221:
```suggestion
        nixpkgs_fmt(package, git_dir)
```

PR #56, opened 2021-05-19
add support for mirror://savannah/

schnusch:
Add a new fetcher that parses [Savannah's](https://download.savannah.nongnu.org/releases/) directory listing.


SuperSandro2000:
@schnusch I think you just need to format the file with black.

Mic92:
Thanks!

PR #60, opened 2021-07-30
Add sourcehut support, minor cleanups

SuperSandro2000:


SuperSandro2000:
not sure why this fails.

```
tests/test_git.py .
error: unrecognised flag '--impure'
Try 'nix --help' for more information.
```

.envrc
```diff
@@ -0,0 +1 @@
+eval "$(lorri direnv)"
```
Mic92, line 1: Please check if lorri is in path (`if command -v lorry >/dev/null`) and fallback to `use nix` otherwise.

SuperSandro2000:
```suggestion
If type -P lorri; then
  eval "$(lorri direnv)"
else
  nix-shell
fi
```

Mic92: More like:

```suggestion
if has lorri; then
  eval "$(lorri direnv)"
else
  use nix
fi
```

`has()` is a direnv builtin.

PR #68, opened 2021-10-08
support for updating yarn FOD hash

yu-re-ka:
For https://github.com/NixOS/nixpkgs/pull/140701


Mic92:
Are you still working on this?

yu-re-ka:
@mic92 @figsoda rebased, added feature note in the readme. this should be ready.

figsoda:
Do you have an example where this works? It doesn't seem to work with `mkYarnPackage` because the `version` attribute always seems to be in [.../yarn2nix-moretea/yarn2nix/default.nix](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/tools/yarn2nix-moretea/yarn2nix/default.nix)

figsoda (approved):
Ah I saw the gotify PR, lgtm

PR #77, opened 2022-01-16
add --write-commit-message for use in scripting

schnusch:


nix_update/__init__.py
```diff
@@ -28,9 +28,15 @@ def parse_args() -> Options:
         "--review", action="store_true", help="Run `nixpkgs-review wip`"
     )
     parser.add_argument("--format", action="store_true", help="Run `nixpkgs-fmt`")
-    parser.add_argument(
+    group = parser.add_mutually_exclusive_group()
```
SuperSandro2000, line 31: I don't think those options need to be exclusive. Sure, it does not make that much sense but it can logically be combined.

nix_update/__init__.py
```diff
@@ -101,6 +108,11 @@ def git_commit(git_dir: str, attribute: str, package: Package) -> None:
             )
 
 
+def write_commit_message(path: str, attribute: str, package: Package) -> None:
+    with open(path, "w") as f:
```
SuperSandro2000, line 112: It would be more fitting if we write this into the cache directory where we already write the report and the two nix files.

schnusch: I think it's more versatile this way. An alternative wood be if nix-update had a formatted stdout, but this would require some work.

PR #94, opened 2022-09-09
Schnusch unstable

Mic92:


nix_update/eval.py
```diff
@@ -65,7 +65,7 @@ def eval_expression(import_path: str, attr: str) -> str:
       line = position.line;
       urls = pkg.src.urls or null;
       url = pkg.src.url or null;
-      rev = pkg.src.url.rev or null;
+      rev = pkg.src.rev or null;
```
zowoq, line 68: Cherry picked this commit to https://github.com/Mic92/nix-update/pull/97

figsoda:
What is the status of this pr? should we contribute to this instead of #76?

nix_update/eval.py
```diff
@@ -59,13 +59,13 @@ def eval_expression(import_path: str, attr: str) -> str:
         builtins.unsafeGetAttrPos "src" pkg;
     in {{
       name = pkg.name;
-      old_version = (builtins.parseDrvName pkg.name).version;
+      old_version = pkg.version or (builtins.parseDrvName pkg.name).version;
```
figsoda, line 62:
```suggestion
      old_version = lib.getVersion pkg;
```

Mic92:
> What is the status of this pr? should we contribute to this instead of #76?

It was a rebased version of that, but it also got stalled. Closing this for now, in case someone wants to pick it up, be my guest.

PR #95, opened 2022-09-11
add vendorHash

zowoq:


zowoq:
> https://github.com/Mic92/nix-update/runs/8299561026
> error: hash '0000000000000000000000000000000000000000000000000000' does not include a type, nor is the type otherwise known from context

@winterqt found another one

zowoq:
The pypi test failure is unrelated, needs to be fixed in `nix-prefetch` in nixpkgs.

README.md
```diff
@@ -10,8 +10,8 @@ designed to work with nixpkgs but also other package sets.
     -   gitlab.com or other instances that uses fetchFromGitLab
     -   pypi
     -   rubygems.org
--   update buildRustPackage's cargoSha256
--   update buildGoModule's vendorSha256/modSha256
+-   update buildRustPackage's cargo{Hash,Sha256}
+-   update buildGoModule's vendor{Hash,Sha256}
```
SuperSandro2000, line 14:
```suggestion
-   update buildRustPackage's cargoHash/cargoSha256
-   update buildGoModule's vendorHash/vendorSha256
```

SuperSandro2000 (commented):
Some changes like ``remove modSha256`` are not obvious why they are done and I first needed to research that this was removed from go modules in 20.09.

README.md
zowoq: done

zowoq:
Seems this works with vendorHash but not vendorSha256 ...

I've cherry picked the readme commit to another PR so once that is merged don't need to keep approving the actions in this PR.

winterqt:
> [Mic92/nix-update/runs/8299561026](https://github.com/Mic92/nix-update/runs/8299561026)
> error: hash '0000000000000000000000000000000000000000000000000000' does not include a type, nor is the type otherwise known from context

@zowoq This is because `fetchPypi` isn't included in `nix-prefetch`'s overlay due to being in the Python package set, so essentially it's looking at the arguments to `fetchurl`, and `fetchPypi` sets `hash` to `""` unconditionally.

This is fixed with the following patch to `nix-prefetch` (applied on top of my existing patch), but I'm not sure if this is the best way to fix this -- special casing for a single fetcher isn't a good idea, IMHO.

```diff
diff --git a/lib/prefetcher.nix b/lib/prefetcher.nix
index fe44a1c..2cd4763 100644
--- a/lib/prefetcher.nix
+++ b/lib/prefetcher.nix
@@ -36,7 +36,7 @@ let
 
   fetcherHashArg =
     let
-      key = if fetcher.args ? hash then "hash" else hashAlgo;
+      key = if fetcher.args ? hash && stringLength fetcher.args.hash != 0 then "hash" else hashAlgo;
     in
     singleAttr key hash;
 ```

The simplest solution here would be to move `fetchPypi` to top-level with all of the other fetchers -- this is probably a good idea anyways since it doesn't do anything dependent on the Python interpreter. @mweinelt @FRidh, as the only people who have touched `fetchPypi` in years: thoughts?

winterqt:
Thinking about this more, adding the `hash != ""` conditional may be the best way to handle this after all.

Will do some more testing to see how this plays out with different circumstances, and report back.

Mic92:
@zowoq added you as a collaborator to this project as well.

SuperSandro2000:
> As the only people who have touched `fetchPypi` in years, what are thoughts on this? I don't see any drawbacks, really -- but I'd appreciate your thoughts before I PR it.

It does not take anything python specific as an input and since python.pkgs inherits pkgs I think it should be save to do. Maybe inherit it into python-packages.nix from pkgs just to be save.

winterqt:
> tests: switch pypi hash to sri

This does fix this issue, but I still think a proper fix in `nix-prefetch` would be good. (Though I really doubt many people are using bare hashes with `fetchPypi`, it's a very isolated issue...)

At least it doesn't block this PR anymore.

(Is there a reason this is a draft?)

zowoq:
> This does fix this issue, but I still think a proper fix in `nix-prefetch` would be good. 

Agreed.

Mic92:
Should we publish a release?

zowoq:
New release would be good, people have started switching to `vendorHash` a bit quicker than I expected.

Mic92:
Done https://github.com/Mic92/nix-update/releases/tag/0.7.0

PR #103, opened 2022-11-14
fix parse_args

figsoda:


winterqt:
When did this break? Seems like it works fine as of latest release.

Mic92:
Here when I was adding changelog support with integration tests: https://github.com/Mic92/nix-update/pull/101/files

PR #104, opened 2022-11-14
remove shebangs in inexecutable files, format with black and isort

figsoda:


zowoq:
https://github.com/Mic92/nix-update/blob/e953da614a4351edb956292a1142eb272b703368/default.nix#L20-L27

Guess this should be added here so it's enforced by CI?

Mic92:
I look into isort in the ci build.

PR #110, opened 2022-11-20
switch to ruff from flake8

Mic92:


default.nix
```diff
@@ -21,13 +20,10 @@ python3.pkgs.buildPythonApplication rec {
   checkPhase = ''
     echo -e "\x1b[32m## run black\x1b[0m"
     LC_ALL=en_US.utf-8 black --check . bin/nix-update
-    echo -e "\x1b[32m## run flake8\x1b[0m"
-    flake8 nix_update
+    echo -e "\x1b[32m## run ruff\x1b[0m"
+    ruff .
```
figsoda, line 24:
```suggestion
    ruff . bin/nix-update
```

default.nix
```diff
@@ -21,13 +20,10 @@ python3.pkgs.buildPythonApplication rec {
...
     echo -e "\x1b[32m## run mypy\x1b[0m"
     mypy --no-warn-unused-ignores --strict nix_update tests
-
-    echo -e "\x1b[32m## run isort\x1b[0m"
-    isort . --check-only
```
figsoda, line 30: `I` or `I001` needs to be added to `select` for ruff to check isort, also there seems to be a small difference in sorting in `nix_update/version/__init__.py` with ruff

Mic92: Ok. No need to rush this. We can wait for the next release of ruff to hit nixpkgs.

figsoda: the ruff update is in nixos-unstable now so it should be good

btw, I updated the ci to use flakes, so you might want to rebase and run `nix flake update --commit-lock-file`

figsoda (approved):

PR #111, opened 2022-11-23
Add outputHash updating

diamondburned:
This commit adds automatic updating of `outputHash`. It doesn't really account for whether `Algo` is set properly, but it works well, I think.

### DIFF
--- a/nix_update/eval.py
+++ b/nix_update/eval.py
@@ -27,6 +27,7 @@ class Package:
     changelog: Optional[str]
     rev: str
     hash: Optional[str]
+    output_hash: Optional[str]
     vendor_hash: Optional[str]
     vendor_sha256: Optional[str]
     cargo_deps: Optional[str]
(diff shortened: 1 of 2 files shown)


Mic92:
Do you have an example where this can be tested?

diamondburned:
I don't really have a particularly easy example, but I needed this for a [Maven derivation that I'm writing](https://github.com/diamondburned/acm-aws/blob/main/packages/crying-counter/default.nix), which relies on `outputHash` to fetch the dependencies.

The file shown here is not updated by this `nix-update` yet, since I'm still experimenting with the repository in a different place. The code seems to work, though.

SuperSandro2000:
Would this work with https://github.com/NixOS/nixpkgs/blob/42aae6fa748a41ced37373fc6d914de512658178/pkgs/development/tools/database/schemaspy/default.nix or https://github.com/NixOS/nixpkgs/blob/42aae6fa748a41ced37373fc6d914de512658178/pkgs/development/tools/database/schemaspy/default.nix ?

diamondburned:
I actually don't know, but I don't think so.

My workaround was to have the `outputHash` derivations be in the global `defaultnix` as well, which is far from ideal (and I wish I can just use `let` like the ones linked above), but I don't think it works there.

If you know of a way for this to work, please let me know and I'll try to implement it. I just can't think of a way that this would work with what `nix-update` is doing.

Mic92 (changes requested):
I also cannot think of a nice heuristics were this would actually work. Since it seems to be rather limited since top-level derivation with output hash and a version that can be bumped are rare, I would not have it for now in nix-update. I am open however if you find a way.

Mic92:
Closing for now since this effort seems to have been stalled.

PR #113, opened 2022-11-24
Add diff feature

figsoda:
Adds a diff url to the commit message similar to the changelog feature  #101, e.g.:
Diff: https://github.com/sharkdp/fd/compare/v8.0.0...v8.5.2
Diff: https://gitlab.gnome.org/world/phosh/phosh/-/compare/v0.20.0...v0.22.0

This integration tests are adapted from test_pypi in #101 as well

### DIFF
--- a/nix_update/eval.py
+++ b/nix_update/eval.py
@@ -1,6 +1,7 @@
 import json
 from dataclasses import InitVar, dataclass, field
 from typing import Any, Dict, List, Optional
+from urllib.parse import ParseResult, urlparse
 
 from .errors import UpdateError
 from .options import Options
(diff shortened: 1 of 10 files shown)


nix_update/eval.py
```diff
@@ -35,10 +37,15 @@ class Package:
 
     raw_version_position: InitVar[Optional[Dict[str, Any]]]
 
+    prased_url: Optional[ParseResult] = None
```
Mic92, line 40:
```suggestion
    parsed_url: Optional[ParseResult] = None
```

figsoda: done

PR #115, opened 2022-11-28
add flag to run passthru.updateScript if possible

figsoda:
currerntly using `-u` with `--commit` doesn't actually commit if the update script only updates files other the original file (e.g. the `pin.json`s), I'm not sure if there is a good way to do it without changing fundamentally how we detect changes

### DIFF
--- a/nix_update/__init__.py
+++ b/nix_update/__init__.py
@@ -32,6 +32,12 @@ def parse_args(args: list[str]) -> Options:
         "--commit", action="store_true", help="Commit the updated package"
     )
     parser.add_argument(
+        "-u",
+        "--use-update-script",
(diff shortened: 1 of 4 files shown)


nix_update/__init__.py
```diff
@@ -31,6 +31,12 @@ def parse_args(args: list[str]) -> Options:
     parser.add_argument(
         "--commit", action="store_true", help="Commit the updated package"
     )
+    parser.add_argument(
+        "-u",
+        "--use-update-script",
```
Mic92, line 36: Can you also mention this in the README?

Mic92:
Works like a charm!

PR #116, opened 2022-11-29
prefer versions with the same prefix

figsoda:
to test, revert cargo-nextest to the 0.9.43 and run `nix-update cargo-nextest`

before: 0.9.43 -> quick-junit-0.3.2
after: 0.9.43 -> 0.9.44


figsoda:
the version regex will apply first, then out of the filtered versions, the ones that match the original prefix will be preferred

PR #118, opened 2022-12-08
Flakes support

### LINKED ISSUE #83: Add support for flakes package output
It would be great if `nix-update` supported updating hashes of packages exposed via the `packages` output of a flake.

figsoda:
fixes #83


Mic92:
That turned out to be easier than anticipated.

PR #125, opened 2022-12-30
Reformat using alejandra

drupol:
This PR:

- [x] Add the `formatter` output attribute
- [x] Reformat nix files using `nix fmt` command

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -4,11 +4,20 @@
   inputs.nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
   inputs.flake-utils.url = "github:numtide/flake-utils";
 
-  outputs = { self, nixpkgs, flake-utils }:
+  outputs = {
+    self,
+    nixpkgs,
+    flake-utils,
+  }:
     flake-utils.lib.eachDefaultSystem (system: {
-      packages.nix-update = nixpkgs.legacyPackages.${system}.callPackage self {
-        src = self;
+      formatter = nixpkgs.legacyPackages.${system}.alejandra;
+
+      packages = {
+        default = self.packages.${system}.nix-update;
+
+        nix-update = nixpkgs.legacyPackages.${system}.callPackage self {
+          src = self;
+        };
       };
-      defaultPackage = self.packages.${system}.nix-update;
     });
 }
--- a/shell.nix
+++ b/shell.nix
@@ -1,9 +1,7 @@
-{ pkgs ? import <nixpkgs> { } }:
-
+{pkgs ? import <nixpkgs> {}}:
 with pkgs;
-
-mkShellNoCC {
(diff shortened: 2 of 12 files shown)


figsoda:
I personally prefer nixpkgs-fmt, but I will leave the decision to @Mic92 

also it is probably best to set the formatter in flake.nix if we want to use a specific formatter

flake.nix
```diff
@@ -4,11 +4,20 @@
...
-  outputs = { self, nixpkgs, flake-utils }:
+  outputs = {
+    self,
+    nixpkgs,
+    flake-utils,
+  }:
     flake-utils.lib.eachDefaultSystem (system: {
-      packages.nix-update = nixpkgs.legacyPackages.${system}.callPackage self {
-        src = self;
+      formatter = nixpkgs.legacyPackages.${system}.alejandra;
```
drupol, line 13: I added the formatter here.

drupol:
> I personally prefer nixpkgs-fmt, but I will leave the decision to @Mic92

It would be so nice to have only one formatter...

flake.nix
figsoda: My bad, I didn't see that at the first glance. edited the comment

SuperSandro2000:
I don't like alejandra either because it is doing almost all things differently on purpose.

tests/testpkgs/savanna.nix
```diff
@@ -1,5 +1,9 @@
-{ lib, stdenv, fetchurl, gamin }:
-
+{
+  lib,
+  stdenv,
+  fetchurl,
+  gamin,
+}:
```
SuperSandro2000, line 6: For example here:
- Why does it enforce multiple lines? Also the commas in the end of the line looks weird.
- Where did the extra new line go?

tests/testpkgs/savanna.nix
```diff
@@ -9,7 +13,7 @@ stdenv.mkDerivation rec {
     sha256 = "0ajc9h023vzpnlqqjli4wbvs0q36nr5p9msc3wzbic8rk687qcxc";
   };
 
-  buildInputs = [ gamin ];
+  buildInputs = [gamin];
```
SuperSandro2000, line 16: Why drop the spaces here? It makes lists more cramped and everything feels much tighter which makes the code harder to read.

shell.nix
```diff
@@ -1,9 +1,7 @@
-{ pkgs ? import <nixpkgs> { } }:
-
+{pkgs ? import <nixpkgs> {}}:
```
SuperSandro2000, line 1: same here, spaces improve readability but alejandra just randomly decided to remove the first and last spaces while keeping the ones around the question mark.

drupol:
Closing in favor of #127

PR #126, opened 2022-12-30
flake: switch from `nixos-unstable` to `nixpkgs-unstable`

drupol:
This PR:

- [x] Switch from `nixos-unstable` to `nixpkgs-unstable`

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,7 +1,7 @@
 {
   description = "Swiss-knife for updating nix packages.";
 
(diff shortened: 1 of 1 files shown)


figsoda:
I'm not against this, but I don't really see the benefits in this change

drupol:
For me it's weird to require `nixos-unstable` in such a package. It should be `nixpkgs`.

But feel free to close the PR if you think it's irrelevant, I'm still a newbie here!

SuperSandro2000:
nixpkgs-unstable has more cache misses. We should stick to nixos-unstable.

PR #128, opened 2023-01-04
Fix ruf error on nixos-unstable

SuperSandro2000:
```
## run black
All done! ✨ 🍰 ✨
28 files would be left unchanged.
## run ruff
error `bin/nix-update` is not supported; Ruff only supports `.py` and `.pyi` files
```


figsoda:
should be fixed in https://github.com/charliermarsh/ruff/pull/1564 (v0.0.208), we should be able to revert this once that gets in nixos-unstable

PR #130, opened 2023-01-05
add --system

### LINKED ISSUE #129: Allow passing system
I have for any system an own fetchurl and want to update them all. It would be nice if I could redo it for an different system

figsoda:
closes #129

### DIFF
--- a/nix_update/__init__.py
+++ b/nix_update/__init__.py
@@ -68,6 +68,11 @@ def parse_args(args: list[str]) -> Options:
         help="Set filename where nix-update will update version/hash",
         default=None,
     )
+    parser.add_argument(
+        "--system",
+        help="The system used to to calculate the hash and run other nix commands",
+        default=None,
+    )
 
     default_attribute = os.getenv("UPDATE_NIX_ATTR_PATH")
     parser.add_argument(
@@ -95,6 +100,7 @@ def parse_args(args: list[str]) -> Options:
         review=a.review,
         format=a.format,
         override_filename=a.override_filename,
+        system_flags=["--system", a.system] if a.system else [],
     )
 
 
@@ -107,7 +113,8 @@ def nix_shell(options: Options) -> None:
                 f"{options.import_path}#{options.attribute}",
                 "--extra-experimental-features",
                 "flakes nix-command",
-            ],
+            ]
+            + options.system_flags,
             stdout=None,
             check=False,
         )
@@ -117,7 +124,7 @@ def nix_shell(options: Options) -> None:
             path = os.path.join(d, "default.nix")
             with open(path, "w") as f:
                 f.write(expr)
-            run(["nix-shell", path], stdout=None, check=False)
+            run(["nix-shell", path] + options.system_flags, stdout=None, check=False)
 
 
 def git_has_diff(git_dir: str, package: Package) -> bool:
@@ -196,7 +203,13 @@ def validate_git_dir(import_path: str) -> str:
 
 
 def nix_run(options: Options) -> None:
-    cmd = ["nix", "shell", "--extra-experimental-features", "flakes nix-command", "-L"]
+    cmd = [
+        "nix",
+        "shell",
+        "--extra-experimental-features",
+        "flakes nix-command",
+        "-L",
+    ] + options.system_flags
 
     if options.flake:
         cmd.append(f"{options.import_path}#{options.attribute}")
@@ -210,7 +223,13 @@ def nix_run(options: Options) -> None:
 
 
 def nix_build(options: Options) -> None:
-    cmd = ["nix", "build", "--extra-experimental-features", "flakes nix-command", "-L"]
+    cmd = [
+        "nix",
+        "build",
+        "--extra-experimental-features",
+        "flakes nix-command",
+        "-L",
+    ] + options.system_flags
     if options.flake:
         cmd.append(f"{options.import_path}#{options.attribute}")
     else:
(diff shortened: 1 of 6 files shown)


nix_update/eval.py
```diff
@@ -117,7 +117,7 @@ def eval_attr(opts: Options) -> Package:
         "nix-command",
         "--expr",
         expr,
-    ]
+    ] + opts.system_flags
```
Mic92, line 120: I don't think this would actually work the way you want. By passing `--system` to nix, it makes the current nix instance believe it is a $system now and attempt to build packages for other operating systems/cpu architectures. The correct way however is setting `localSystem` when importing nixpkgs (not needed for flakes as this is already correctly set when exposing packages). This will make sure that a remote builder is used whenever a package needs to be build that does not match the current system.

nix_update/__init__.py
```diff
@@ -196,7 +203,13 @@ def validate_git_dir(import_path: str) -> str:
 
 
 def nix_run(options: Options) -> None:
-    cmd = ["nix", "shell", "--extra-experimental-features", "flakes nix-command", "-L"]
+    cmd = [
+        "nix",
+        "shell",
+        "--extra-experimental-features",
+        "flakes nix-command",
+        "-L",
```
SuperSandro2000, line 211: We should deduplicate the arguments

nix_update/eval.py
figsoda: Can you elaborate on the differences? pretty sure `localSystem` [defaults](https://github.com/NixOS/nixpkgs/blob/9813adc7f7c0edd738c6bdd8431439688bb0cb3d/pkgs/top-level/impure.nix#L17) to `builtins.currentSystem`, which is [equilvalent](https://github.com/NixOS/nix/blob/d02c5a41da68908bd88d7a697abc52a04826894e/src/libexpr/primops.cc#L3984-L3985) to `--system` [option](https://github.com/NixOS/nix/blob/d02c5a41da68908bd88d7a697abc52a04826894e/src/libstore/globals.hh#L157)

nix_update/__init__.py
figsoda: by that do you mean adding `--extra-experimental-features "flakes nix-command"` unconditionally to `system_flags` (maybe rename it to extra_flags)?

SuperSandro2000: or just move the list to a common variable and re-use that all over. or what you suggested if we use it everywhere.

nix_update/eval.py
Mic92: localSystem defaults to builtins.currentSystem. However changing it to a different value is not the same as changing the system of the locally running nix. When you change the system of nix it tries to build aarch64-linux binaries on an x86-64-linux:

```
$ nix-build --system 'aarch64-linux' --builders '' -E "with import <nixpkgs> {}; pkgs.writeText \"foo\" ''Test''"
ErrorErrorvexecuting '/nix/store/hjala5j03vxhkj6rvgj8kdigm1yrw0p4-bash-5.1-p16/bin/bash': Exec format error
error: builder for '/nix/store/m16gsw4qlb6l2hqqnsb7dy74kldbv54i-foo.drv' failed with exit code 1;
       last 2 log lines:
       > 
       > ErrorErrorvexecuting '/nix/store/hjala5j03vxhkj6rvgj8kdigm1yrw0p4-bash-5.1-p16/bin/bash': Exec format error
       For full logs, run 'nix log /nix/store/m16gsw4qlb6l2hqqnsb7dy74kldbv54i-foo.drv'.
```

Whereas if you change system than it will use a remote builder if available:

```
$  nix-build --builders '' -E "with import <nixpkgs> { system = \"aarch64-linux\"; }; pkgs.writeText \"foo\" ''Test''"
this derivation will be built:
  /nix/store/m16gsw4qlb6l2hqqnsb7dy74kldbv54i-foo.drv
error: a 'aarch64-linux' with features {} is required to build '/nix/store/m16gsw4qlb6l2hqqnsb7dy74kldbv54i-foo.drv', but I am a 'x86_64-linux' with features {benchmark, big-parallel, gccarch-broadwell, gccarch-haswell, gccarch-ivybridge, gccarch-san
dybridge, gccarch-skylake, gccarch-westmere, gccarch-znver1, gccarch-znver2, gccarch-znver3, kvm, nixos-test}
```

figsoda: I can't reproduce that, but anyways, I have added that to the import inputs
```
$ nix-build --system 'aarch64-linux' --builders '' -E "with import <nixpkgs> {}; pkgs.writeText \"foo\" ''Test''"
this derivation will be built:
  /nix/store/q0x61xkb7139ksc9dn3sc9n26sr0nbai-foo.drv
error: a 'aarch64-linux' with features {} is required to build '/nix/store/q0x61xkb7139ksc9dn3sc9n26sr0nbai-foo.drv', but I am a 'x86_64-linux' with features {benchmark, big-parallel, kvm, nixos-test}
```

nix_update/eval.py
```diff
@@ -54,22 +54,28 @@ def __post_init__(self, raw_version_position: Optional[Dict[str, Any]]) -> None:
...
-          pos // {{ file = "{import_path}" + substring outPathLen (stringLength file - outPathLen) file; }};
+            assert substring 0 outPathLen file == outPath;
+            pos // {{ file = "{import_path}" + substring outPathLen (stringLength file - outPathLen) file; }};
         """
     else:
         let_bindings = f"""
           inputs = if (builtins.functionArgs (import {import_path})) ? overlays then {{ overlays = [ ]; }} else {{ }};
-          pkg = (import {import_path} inputs).{attr};
+          system = if (builtins.functionArgs (import {import_path})) ? system then {{ system = {system}; }} else {{ }};
+          pkg = (import {import_path} (inputs // system)).{attr};
```
Mic92, line 78: This makes it a bit easier to follow the code:

```  
        let_bindings = f"""
          packageSet = import {import_path};
          args =   builtins.functionArgs pkgs;
          inputs = (if args & system then {{ inherit system; }} else {{}})  //
                        (if args & overlays then {{ inherit overlays; }} else {{}});
          pkg = (packageSet inputs).{attr};
        """
```

Mic92 (approved):
Looks good to me when CI also gets green.

nix_update/__init__.py
```diff
@@ -68,6 +68,11 @@ def parse_args(args: list[str]) -> Options:
         help="Set filename where nix-update will update version/hash",
         default=None,
     )
+    parser.add_argument(
+        "--system",
+        help="The system used to to calculate the hash and run other nix commands",
+        default=None,
```
Mic92, line 74: Is this better?
 
```suggestion
        default="builtins.currentSystem",
```

figsoda: this wouldn't work when passed to `--system`, also things like `aarch64-linux` needs to be quoted

PR #143, opened 2023-03-21
add support for updating packages using importCargoLock

figsoda:
this raises the minimum supported python version to 3.11 and also adds nix-prefetch-git as a runtime dependency

related: https://github.com/NixOS/nixpkgs/pull/217084, https://github.com/NixOS/nixpkgs/pull/221716


nix_update/update.py
```diff
@@ -131,6 +147,81 @@ def update_cargo_deps_hash(opts: Options, filename: str, current_hash: str) -> N
     replace_hash(filename, current_hash, target_hash)
 
 
+def update_cargo_lock(opts: Options, filename: str, dst: str) -> None:
```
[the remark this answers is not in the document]
figsoda: added one test for each of these two
https://github.com/mic92/nix-update/blob/4bde7565ab4b18d42f467f0197364557452f7bf1/nix_update/update.py#L185
https://github.com/mic92/nix-update/blob/4bde7565ab4b18d42f467f0197364557452f7bf1/nix_update/update.py#L195

pyproject.toml
```diff
@@ -1,11 +1,12 @@
 [tool.ruff]
+target-version = "py311"
 line-length = 88
 
 select = ["E", "F", "I"]
 ignore = [ "E501" ]
 
 [tool.mypy]
-python_version = "3.10"
+python_version = "3.11"
```
SuperSandro2000, line 9: Please keep in mind that python 3.11 support in nixpkgs can still be spotty especially with some slower updating python packages.

figsoda: Thought this might be better than introducing an external dependency, since parsing toml is pretty much required for this

Mic92: I don't think this includes the python 3.11 interpreter itself. nix-update does not use third-party dependencies.

PR #148, opened 2023-04-03
fix updating `importCargoLock` when `lockFile` is in the nix store

### LINKED ISSUE #147: Can't update cargo package when lock file references the src's lock file
As far as I can tell, I can't make `nix-update` work whenever I use this:

```
  cargoLock = {
    lockFile = "${src.out}/Cargo.lock";
    allowBuiltinFetchGit = true;
  }
```

error:

```
error: assertion '((substring 0 outPathLen file) == outPath)' failed
```

figsoda:
closes #147


colemickens:
This works for my nushell+wezterm derivations pointed at my fork. Thank you very much!

nix_update/update.py
```diff
@@ -152,25 +152,38 @@ def update_cargo_deps_hash(opts: Options, filename: str, current_hash: str) -> N
     replace_hash(filename, current_hash, target_hash)
 
 
-def update_cargo_lock(opts: Options, filename: str, dst: str) -> None:
+def update_cargo_lock(opts: Options, filename: str, dst: CargoLockInSource | CargoLockInStore) -> None:
```
Mic92, line 155: I found the previous type a bit hard to read so I introduced a new more descriptive type.

nix_update/eval.py
```diff
@@ -33,26 +52,41 @@ class Package:
...
             self.parsed_url = urlparse(url)
         if raw_version_position is None:
             self.version_position = None
         else:
             self.version_position = Position(**raw_version_position)
+            raw_cargo_lock
+        if raw_cargo_lock is None:
+            self.cargo_lock = NoCargoLock()
+        elif raw_cargo_lock is False:
+            self.cargo_lock = CargoLockInStore()
```
Mic92, line 85: I am unsure about this part. If we get false, do we want to throw an error instead?

figsoda: this is false when it is a store path in a flake, so I would say the current solution is good

figsoda:
ofborg merge

figsoda:
bors merge

PR #150, opened 2023-04-09
switch to go-modules.outputHash

zowoq:


figsoda:
we should probably keep this around for a little longer because people do use nix-update outside nixpkgs

zowoq:
> we should probably keep this around for a little longer because people do use nix-update outside nixpkgs

This'll be necessary as otherwise nix-update won't work with nixpkgs as vendorSha256 == vendorHash.

We've already broken prior compatibility once when we added vendorHash and _unset.

This PR should actually restore that compatibility but I don't think it's worth the hassle to trying to keep this around as well.

figsoda:
> This'll be necessary as otherwise nix-update won't work with nixpkgs as vendorSha256 == vendorHash.

Understandable then :) I wasn't aware of that

Would using `go-modules.outputHash` perhaps be more compatible?

zowoq:
> Would using `go-modules.outputHash` perhaps be more compatible?

iirc it didn't work when vendorHash was first added here but it does work now.

figsoda (approved):

PR #155, opened 2023-04-20
improve assertion message

### LINKED ISSUE #146: trace some variables to make debugging easier
Hi.

I'm trying to use nix-update to update `nushell` to the tip-of-tree of my fork. I got it to work once, but now it seems to fail:

```
error: assertion '((substring 0 outPathLen file) == outPath)' failed
```

It would be helpful if outPath/file were traced so I could have a better idea of what's going on

figsoda:
closes #146


Mic92:
bors merge

PR #157, opened 2023-04-27
fix updating cargoLock with flakes

### LINKED ISSUE #153: nix-update fails on cargo packages with git deps
```
error: A hash was specified for reedline-0.18.0, but there is no corresponding git dependency.
```

I just hit this when updating `nushell` again. It seems nix-update doesn't handle removed git cargo deps?

figsoda:
hopefully fixes #153 

using `toString` stops `Cargo.lock` from being copied to the store alone, and makes sure it is still under the directory of the flake, so nix-update is able to find its local path


colemickens:
worked great, thank you!

Issue #170, opened 2023-06-01
commit changed files directly rather than `git add` them


colemickens:
Why does the code alter the staging area when it seems to commit the directly changed files at the end anyway?

PR #171, opened 2023-06-01
git-commit files directly without using staging area

Mic92:
fixes https://github.com/Mic92/nix-update/issues/170


figsoda (approved):
bors r+

PR #179, opened 2023-08-01
escape attribute names

### LINKED ISSUE #178: Slashes in the name of an attribute result in a parse error, and should be put in quotes
Consider the following example (which I'm afraid I can't release the code to), where I want to update an attribute under the name `psql_15/exts/pg_hashids` &mdash; traceback and some fluff omitted:

```
$ nix-update --flake --version=branch psql_15/exts/pg_hashids

...
error: undefined variable 'or'

       at «string»:7:40:

            6|   flake = getFlake "/home/austin/work/nix-postgres";
            7|   pkg = flake.packages.${currentSystem}.psql_15/exts/pg_hashids or flake.psql_15/exts/pg_hashids;
             |                                        ^
            8|   inherit (flake) outPath;
...
```

The error'ing line itself is pretty clear:

```
pkg = flake.packages.${currentSystem}.psql_15/exts/pg_hashids or flake.psql_15/exts/pg_hashids;
```

The name is generated by way of `flattenTree` from `flake-utils` so, it isn't exactly a super strange case IMO. I think just enclosing the name in general is enough if there's a slash. Really, we could just do this unconditionally, since there isn't any real difference between `x.y` and `x."y"` other than variable interpolation?

The following command works, in contrast, but is obviously strange:

`nix-update --flake --version=branch \"psql_15/exts/pg_hashids\"`

figsoda:
closes #178

it uses `json.dumps` which may not be the most ideal, but it is the easiest way I can think of

### DIFF
--- a/nix_update/options.py
+++ b/nix_update/options.py
@@ -1,3 +1,4 @@
+import json
 import os
 from dataclasses import dataclass, field
 from typing import List, Optional
@@ -26,3 +27,6 @@ class Options:
     format: bool = False
     system: Optional[str] = None
     extra_flags: List[str] = field(default_factory=list)
+
+    def __post_init__(self) -> None:
+        self.attribute = json.dumps(self.attribute)


thoughtpolice (approved):
wfm 👍🏻

figsoda:
this unfortunately caused a few bugs which should be fixed in #180, that's why I'm skipping the 0.19.1 update in nixpkgs
i should have done a bit more testing

PR #199, opened 2023-09-30
Modernize buildsystem

Mic92:


default.nix
```diff
@@ -3,13 +3,16 @@
 
 
 pkgs.python311.pkgs.buildPythonApplication {
-  name = "nix-update";
+  pname = "nix-update";
+  version = "0.19.3";
```
SuperSandro2000, line 7: we could probably import pyproject.toml and read the version in code from there

PR #203, opened 2023-10-21
Add support for updating buildDartApplication's vendorHash

lelgenio:
⚠️ I'm unfamiliar with this code base, so I may have missed adding support for something.

I had to get creative to get the current vendor hash. It's unfortunately not exported as part of `package.dartDeps`.


Mic92:
You can probably fix upstream to export `partDeps` as it is done for the rest.

Mic92:
Just re-open the pull request if you continue working on this.

Issue #177, opened 2023-07-31
Switch from `cargoSha256` to `cargoHash`


asymmetric:
If a package has a SRI hash, then the manual [recommends](https://nixos.org/manual/nixpkgs/stable/#compiling-rust-applications-with-cargo) using `cargoHash`.

It could be handy if `nix-update` could handle this, to avoid small annoyances like https://github.com/NixOS/nixpkgs/pull/246037#pullrequestreview-1553423677.

Mic92:
We are currently don't have robust enough mechanism in nix-update to perform refactorings like that: https://github.com/Mic92/nix-update/pull/209

PR #209, opened 2023-11-25
Change hash attr to algo agnostic one (sha256 -> hash, cargoSha256 -> cargoHash)

Kranzes:
As requested previously in https://github.com/Mic92/nix-update/issues/177, this change can help with minimizing the nitpicks on new nixpkgs PRs because of the new fetchers' hash attribute name conventions.

This works for:
 - `sha256` -> `hash`
 - `cargoSha256` -> `cargoHash`
 - `vendorSha256` -> `vendorHash`
 - and more

Note: It currently only supports `sha256` hashes as I saw that in a couple of lines below we only get the hash for `sha256` anyway, also, the other algorithms aren't that popular anyway.

### DIFF
--- a/nix_update/update.py
+++ b/nix_update/update.py
@@ -78,6 +78,9 @@ def replace_hash(filename: str, current: str, target: str) -> None:
         with fileinput.FileInput(filename, inplace=True) as f:
             for line in f:
                 line = line.replace(current, normalized_hash)
+                # Also update hash attr name
+                line = re.sub(r"sha256\s*=\s*", r"hash = ", line)
+                line = re.sub(r"Sha256\s*=\s*", r"Hash = ", line)
                 print(line, end="")
 
 


Kranzes:
Do I need to change anything in the tests?

Mic92:
I don't think I want nix-update to also refactor the code. Replacing hashes is somewhat reliable because they are practically unique but replacing random `sha256 = ` variables could replace instances it shouldn't replace. I think for that it's better to have a tool that understand the language and picks up what a fetcher call is rather just doing search and replace.

Kranzes:
Fair. I guess while at it also close the issue about this? Btw, isn't `--format` refactoring too in some sense?

Mic92:
Yes but `--format` is using a reliable nix formatter that has support for nix syntax on AST level rather than just doing substitution. Here an example:

```
  some-package = some-generic-builder {
    sha256 = "foo";
  };
```

We sometimes have builder functions i.e. postgres where we maintain multiple versions of the same program. Those function may not accept hash but only sha256 as an input and would be broken as a result of this. Even if we fix this in every instance in nixpkgs there are still third-party repos that also can use nix-update.

Mic92:
You can just make yourself a wrapper that uses sed on all git-modified files to do the replacement.

PR #214, opened 2024-01-09
Add option for version_key

pineapplehunter:
Hi, I was having some problems with packaging with different versioning, so I copied this idea from `common-updater-scripts` from nixpkgs.

Some packages can have a different version from the source repo but could be derived from source. Currently I see no way to update these packages like this with `nix-update`.
This PR adds an option `version_key` for this case.
By expose a variable that holds the github repo version, `nix-update *name* --version-key *version attr name*` could update the package.

BTW, I added an option to the argparser so it shows the default values of options on `--help`.

Thanks!

### DIFF
--- a/nix_update/__init__.py
+++ b/nix_update/__init__.py
@@ -18,7 +18,9 @@ def die(msg: str) -> NoReturn:
 
 
 def parse_args(args: list[str]) -> Options:
-    parser = argparse.ArgumentParser()
+    parser = argparse.ArgumentParser(
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter
+    )
     help = "File to import rather than default.nix. Examples, ./release.nix"
     parser.add_argument("-f", "--file", default="./.", help=help)
     parser.add_argument(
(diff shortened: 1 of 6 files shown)


pineapplehunter:
I just leant that the case I was thinking of (in the test case) was using import from derivation.
I don't have any other examples that use this feature, so maybe this should just be closed.

nix_update/__init__.py
```diff
@@ -56,6 +58,12 @@ def parse_args(args: list[str]) -> Options:
         help="Regex to extract version with, i.e. 'jq-(.*)'",
         default="(.*)",
     )
+    parser.add_argument(
+        "-vk",
+        "--version-key",
```
Mic92, line 63: I think version-attr fits better with how nix language names things:

```suggestion
        "--version-attr",
```

nix_update/__init__.py
```diff
@@ -56,6 +58,12 @@ def parse_args(args: list[str]) -> Options:
...
+        help="Key of attribute that holds the version",
```
Mic92, line 64:
```suggestion
        help="Name of attribute that holds the version",
```

nix_update/__init__.py
```diff
@@ -56,6 +58,12 @@ def parse_args(args: list[str]) -> Options:
         help="Regex to extract version with, i.e. 'jq-(.*)'",
         default="(.*)",
     )
+    parser.add_argument(
+        "-vk",
```
Mic92, line 62: Not sure we need a short-hand for that:

```suggestion
```

I see this more being used in scripts where the long-form is more descriptive.

Mic92:
Yeah. Maybe we can revisit this when we have a better idea how this would be useful...
Most packages will have version set properly.

PR #219, opened 2024-01-17
make gitlab updater work by tag when releases are disabled

lilyinstarlight:
If a GitLab repository has the "Releases" feature disabled (such as [vcard](https://gitlab.com/engmark/vcard)), the updater currently fails with `KeyError: 'release'` since the release key is removed from the API

This allows it to gracefully fall back to using tags even if releases are disabled


l0b0 (approved):

l0b0:
@Mic92 Would you be OK to do a release for `nixpkgs/pkgs/tools/package-management/nix-update/default.nix b/pkgs/tools/package-management/nix-update/default.nix`?

PR #221, opened 2024-01-19
fix: detect prereleases on GitHub so we can ignore them

lf-:
This is simple enough: we just use the GitHub releases API that does actually have the info in it.

Uses this GitHub API which is fine to use unauthenticated: https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#list-releases

I have complained about the feed not having the info here: https://github.com/orgs/community/discussions/88765

### DIFF
--- a/nix_update/version/github.py
+++ b/nix_update/version/github.py
@@ -1,23 +1,20 @@
+import json
 import re
 import urllib.request
 import xml.etree.ElementTree as ET
-from urllib.parse import ParseResult, unquote, urlparse
-from xml.etree.ElementTree import Element
+from urllib.parse import ParseResult, urlparse
 
-from ..errors import VersionError
 from ..utils import info
 from .version import Version
 
 
-def version_from_entry(entry: Element) -> Version:
-    if entry is None:
-        raise VersionError("No release found")
-    link = entry.find("{http://www.w3.org/2005/Atom}link")
-    assert link is not None
-    href = link.attrib["href"]
-    url = urlparse(href)
-    # TODO: set pre-release flag
-    return Version(unquote(url.path.split("/")[-1]))
+def version_from_entry(entry: dict) -> Version:
+    href = entry["html_url"]
+    assert href is not None
+    prerelease = entry["prerelease"]
+    draft = entry["draft"]
+    version = entry["tag_name"]
+    return Version(version, prerelease=prerelease or draft)
 
 
 def fetch_github_versions(url: ParseResult) -> list[Version]:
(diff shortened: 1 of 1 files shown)


nix_update/version/github.py
```diff
@@ -27,11 +24,19 @@ def fetch_github_versions(url: ParseResult) -> list[Version]:
     owner, repo = parts[1], parts[2]
     repo = re.sub(r"\.git$", "", repo)
     # TODO fallback to tags?
-    feed_url = f"https://github.com/{owner}/{repo}/releases.atom"
-    info(f"fetch {feed_url}")
-    resp = urllib.request.urlopen(feed_url)
-    tree = ET.fromstring(resp.read())
-    releases = tree.findall(".//{http://www.w3.org/2005/Atom}entry")
+    # https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#list-releases
+    releases_url = f"https://api.github.com/repos/{owner}/{repo}/releases?per_page=100"
```
Mic92, line 28: If we switch to the API we also need to accept GITHUB_TOKENS (potentially optional).
Otherwise people behind shared IPs will run into API limits easily.

Mic92: Also does this detect git tags? Not all projects use github releases

lf-: Damn, it doesn't, and that's really annoying.

lf-: As far as I can tell, we would have to use graphql to get commit info (or even date information) next to tags, and there's no way to get combined data between releases and tags aside from the very clever thing we're currently doing. This sucks!

lf-: Of course:

```graphql
query ($owner: String!, $name: String!, $curs: String!) {
  repository(owner: $owner, name: $name) {
    refs(
      refPrefix: "refs/tags/"
      after: $curs
      first: 100
      orderBy: {field: TAG_COMMIT_DATE, direction: DESC}
    ) {
      nodes {
        id
        name
        target {
          oid
          __typename
          ... on Commit {
            author {
              name
              date
            }
          }
        }
      }
    }
  }
}
```

```json
{
  "data": {
    "repository": {
      "refs": {
        "nodes": [
          {
            "id": "MDM6UmVmMjgwMTAyODYyOnJlZnMvdGFncy92MC42LjU=",
            "name": "v0.6.5",
            "target": {
              "oid": "f0c17fdad5de2ae2d5893943bc437d56e459ea2e",
              "__typename": "Commit",
              "author": {
                "name": "Jade Lovelace",
                "date": "2024-02-03T21:34:44.000-08:00"
              }
            }
          },
...
```

But I don't like this, among other reasons, because I believe it would mandate having github tokens always.

PR #229, opened 2024-03-01
make cargo lock updater keep outpath links in temporary directory

lilyinstarlight:
This fixes running `nix-update` on cargo lock packages leaving result links laying around

### DIFF
--- a/nix_update/update.py
+++ b/nix_update/update.py
@@ -163,14 +163,17 @@ def update_cargo_deps_hash(opts: Options, filename: str, current_hash: str) -> N
 def update_cargo_lock(
     opts: Options, filename: str, dst: CargoLockInSource | CargoLockInStore
 ) -> None:
-    res = run(
-        [
-            "nix",
-            "build",
-            "--impure",
-            "--print-out-paths",
-            "--expr",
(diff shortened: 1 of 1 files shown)


nix_update/update.py
```diff
@@ -163,10 +163,14 @@ def update_cargo_deps_hash(opts: Options, filename: str, current_hash: str) -> N
 def update_cargo_lock(
     opts: Options, filename: str, dst: CargoLockInSource | CargoLockInStore
 ) -> None:
+    tempdir = tempfile.TemporaryDirectory()
```
Mic92, line 166: In the event of a build failure this would not cleanup `tempdir` unless `with tempdir:` is used. So we need to `--no-out-link` to `nix build` to skip the symlink all together or add a large with statement.

lilyinstarlight: I considered both options and `--no-out-link` made me worry about GC races, but I am fine to add the `with` statement (it was a bigger diff and it was already leaving them everywhere, so I didn't initially do that)

I'll push the `with tempfile.TemporaryDirectory() as tempfile:` version sometime in the next hour or two

Thank you!

Issue #238, opened 2024-03-28
updating yarn offlineCache fails: `.../yarn2nix/default.nix is not in /nix/store/...-source`


katexochen:
I'm trying to update the `offlineCache` of a package that uses `mkYarnPackage` as builder, but I can't get nix-update to find the right attribute position of `src`. Any suggestions?

Code is here: https://github.com/katexochen/docusaurus-testing/tree/nix-update

```
nix-update --version=skip --override-filename=flake.nix --flake default 
$ nix eval --json --impure --expr 
let
  
  inherit (builtins) getFlake stringLength substring;
  currentSystem = builtins.currentSystem;
  flake = getFlake "/home/katexochen/katexochen/docusaurus-testing";
  pkg = flake.packages.${currentSystem}."default" or flake."default";
  inherit (flake) outPath;
  outPathLen = stringLength outPath;
  sanitizePosition = { file, ... }@pos:
    assert substring 0 outPathLen file != outPath
      -> throw "${file} is not in ${outPath}";
    pos // { file = "/home/katexochen/katexochen/docusaurus-testing" + substring outPathLen (stringLength file - outPathLen) file; };

  raw_version_position = sanitizePosition (builtins.unsafeGetAttrPos "version" pkg);

  position = if pkg ? isRubyGem then
    raw_version_position
  else if pkg ? isPhpExtension then
    raw_version_position
   else
    sanitizePosition (builtins.unsafeGetAttrPos "src" pkg);
in {
  name = pkg.name;
  old_version = pkg.version or (builtins.parseDrvName pkg.name).version;
  inherit raw_version_position;
  filename = position.file;
  line = position.line;
  urls = pkg.src.urls or null;
  url = pkg.src.url or null;
  rev = pkg.src.rev or null;
  hash = pkg.src.outputHash or null;
  go_modules = pkg.goModules.outputHash or null;
  go_modules_old = pkg.go-modules.outputHash or null;
  cargo_deps = pkg.cargoDeps.outputHash or null;
  raw_cargo_lock =
    if pkg ? cargoDeps.lockFile then
      let
        inherit (pkg.cargoDeps) lockFile;
        res = builtins.tryEval (sanitizePosition {
          file = toString lockFile;
        });
      in
      if res.success then res.value.file else false
    else
      null;
  composer_deps = pkg.composerRepository.outputHash or null;
  npm_deps = pkg.npmDeps.outputHash or null;
  yarn_deps = pkg.offlineCache.outputHash or null;
[...]

PR #239, opened 2024-04-01
respect --override-filename flag for flakes

daniel-sampliner:
fixes https://github.com/Mic92/nix-update/issues/238


Mic92:
@mergify queue

PR #241, opened 2024-04-12
move bin/nix-update to nix_update/__main__.py

lucasew:
Behaviour or $out shouldn't change. It's basically just a Python trick.


Mic92:
That works technically but is a bin folder that has the matching executable, not nicer?
i.e. this enables `export PATH=/path/to/nix-update/bin:$PATH`

Mic92:
I prefer the current version.

PR #244, opened 2024-04-20
Remove wrongly advertised support for notabug.com, make `fetch_gitea_snapshots` less expensive and fix `fetchFromGitea`'s fallback behaviour to `fetchgit` 

emilylange:
Please refer to the commit messages for details.

Note: `fetchFromGitLab` has the same `fetchgit` fallback behavior and as such it is also affected by the same issue.


Mic92:
@mergify queue

PR #246, opened 2024-04-28
github: Use API to properly tag prereleases

jvanbruegge:
r-ryantm keeps opening pull requests for a github package where the latest release is a prerelease
https://github.com/NixOS/nixpkgs/pulls?q=is%3Apr+is%3Aclosed+audible-cli

The reason is that `nix-updated` uses the RSS feed to get the releases which does not distinguish between a release and a prerelease. Using the github API fixes that.


tests/conftest.py
```diff
@@ -33,6 +33,10 @@ def testpkgs(init_git: bool = False) -> Iterator[Path]:
 
                 subprocess.run(["git", "-C", tmpdirname, "init"], check=True)
                 subprocess.run(["git", "-C", tmpdirname, "add", "--all"], check=True)
+                subprocess.run(
+                    ["git", "-C", tmpdirname, "config", "commit.gpgsign", "false"],
+                    check=True,
+                )
```
jvanbruegge, line 39: Without this the tests cannot be run by someone that has enabled gpgsign globally

nix_update/version/github.py
```diff
@@ -1,38 +1,25 @@
...
     owner, repo = parts[1], parts[2]
     repo = re.sub(r"\.git$", "", repo)
     # TODO fallback to tags?
-    feed_url = f"https://github.com/{owner}/{repo}/releases.atom"
-    info(f"fetch {feed_url}")
-    resp = urllib.request.urlopen(feed_url)
-    tree = ET.fromstring(resp.read())
-    releases = tree.findall(".//{http://www.w3.org/2005/Atom}entry")
-    return [version_from_entry(x) for x in releases]
+    github_url = f"https://api.github.com/repos/{owner}/{repo}/releases"
```
Mic92, line 18: When using the github api, we would also need to support using a github token i.e. through environment variables. Otherwise it might fail to fetch updates i.e. when users are behind a shared ip or in CI like github because of rate limits.

I would suggest to keep both the RSS feed and also use the API if a GITHUB_TOKEN is provided.

Mic92: @jvanbruegge what do you think?

jvanbruegge: Yes, I can do it. Just theblast few weeks were quite stressfull and I will probably not get to it before the end of next week

jvanbruegge:
@Mic92 I've added the atom feed as a fallback now.

jvanbruegge:
The test failure is from `pnpm` and it also happens on master, so is unrelated to this PR

Mic92:
@mergify queue

jvanbruegge:
Could you cut a new release? I'd really like to get rid of the r-ryantm spam

Mic92:
Release is out.

Mic92:
There is one issue with this. Some projects don't have github releases i.e. https://github.com/mongodb/mongo/

On the other hand tags work: https://api.github.com/repos/mongodb/mongo/tags?per_page=1000

SebTM:
> There is one issue with this. Some projects don't have github releases i.e. https://github.com/mongodb/mongo/

Yes, ran into this issue for protonvpn-gui and dependencies today:

https://api.github.com/repos/ProtonVPN/proton-vpn-gtk-app/releases

does not contain the tags but the old atom-feed did. Can we have a switch for the new behavior and keep the old default for now?

jvanbruegge:
I can add a second fallback in case list of releases is empty. I think that is a better approach

jvanbruegge:
@Mic92 @SebTM see #269

lilyinstarlight:
This breaks so many packages that have used GitHub releases in the past but only push tags these days. The audible-cli package should be changed to use `--version-regex` to fix its issue (as is already done in nixpkgs packages with this issue, such as [aws-sam-cli](https://github.com/NixOS/nixpkgs/blob/71e91c409d1e654808b2621f28a327acfdad8dc2/pkgs/development/tools/aws-sam-cli/default.nix#L145)) and this be reverted in nix-update rather than breaking updates for a lot of packages

lf-:
https://github.com/Mic92/nix-update/pull/221 related pr, closed because it would have been a regression

jian-lin:
> I can add a second fallback in case list of releases is empty.

I think this is not enough.  Some packages, such as [kubernetes] and [lspce], have a non-empty release list but there are newer tags which are not included in the release list.

[kubernetes]: https://github.com/Mic92/nix-update/issues/272
[lspce]: https://github.com/NixOS/nixpkgs/pull/340005

lilyinstarlight:
Partial revert in #274

PR #247, opened 2024-04-30
 Try pkg.meta.position before unsafeGetAttrPos

pbsds:
Some package builders set `meta.position` manually due to clobbering the origin of `src`.

Fixes `nix-update -u mpvScripts.*`, discovered while reviewing https://github.com/NixOS/nixpkgs/pull/308062.

### DIFF
--- a/nix_update/eval.py
+++ b/nix_update/eval.py
@@ -142,13 +142,22 @@ def eval_expression(
     return f"""
 let
   {indent(dedent(let_bindings), "  ")}
+  positionFromMeta = pkg: let
+    parts = builtins.match "(.*):([0-9]+)" pkg.meta.position;
+  in {{
+    file = builtins.elemAt parts 0;
+    line = builtins.fromJSON (builtins.elemAt parts 1);
+  }};
+
   raw_version_position = sanitizePosition (builtins.unsafeGetAttrPos "version" pkg);
 
-  position = if pkg ? isRubyGem then
+  position = if pkg ? meta.position then
+    sanitizePosition (positionFromMeta pkg)
+  else if pkg ? isRubyGem then
     raw_version_position
   else if pkg ? isPhpExtension then
     raw_version_position
-   else
+  else
     sanitizePosition (builtins.unsafeGetAttrPos "src" pkg);
 in {{
   name = pkg.name;


nix_update/eval.py
```diff
@@ -142,14 +142,23 @@ def eval_expression(
...
   position = if pkg ? isRubyGem then
     raw_version_position
   else if pkg ? isPhpExtension then
     raw_version_position
-   else
-    sanitizePosition (builtins.unsafeGetAttrPos "src" pkg);
+  else if (builtins.unsafeGetAttrPos "src" pkg) != null then
+    sanitizePosition (builtins.unsafeGetAttrPos "src" pkg)
+  else
+    sanitizePosition (positionFromMeta pkg);
```
Mic92, line 161: If we have meta.position, should it not be preferred over `src`?

pbsds: i was trying to change as little of existing behaviour as possible, by if you're on board with this change then why not.

nix_update/eval.py
```diff
@@ -142,13 +142,22 @@ def eval_expression(
     return f"""
 let
   {indent(dedent(let_bindings), "  ")}
+  positionFromMeta = pkg: let
+    parts = builtins.match "(.*):([0-9]+)" pkg.meta.position;
+  in if pkg ? meta.position then {{
```
Mic92, line 147: Do we need this `if` here? It's already checked on line 158. Also would make the code slightly simpler.

nix_update/eval.py
```diff
@@ -142,13 +142,22 @@ def eval_expression(
...
+
   raw_version_position = sanitizePosition (builtins.unsafeGetAttrPos "version" pkg);
 
   position = if pkg ? isRubyGem then
     raw_version_position
   else if pkg ? isPhpExtension then
     raw_version_position
-   else
+  else if pkg ? meta.position then
+    sanitizePosition (positionFromMeta pkg)
```
Mic92, line 159: I think we can even overrule the quirks for ruby gems and php here, I would say.

Mic92:
@mergify queue

doronbehar (commented):
This PR broke my updating workflow for `gnuradioMinimal` :( I don't set manually `meta.position`, but I did set `src` in the file where the hash and version are actually written... Also, setting the `meta.position` manually doesn't work for me from some reason.

Mic92:
what is meta.position set to in this case?

doronbehar:
To be more specific, here's my (Nixpkgs) patching attempt of trying to set the `meta.position` manually:

```diff
diff --git i/pkgs/applications/radio/gnuradio/wrapper.nix w/pkgs/applications/radio/gnuradio/wrapper.nix
index f468aadb27b5..107efe90bc80 100644
--- i/pkgs/applications/radio/gnuradio/wrapper.nix
+++ w/pkgs/applications/radio/gnuradio/wrapper.nix
@@ -163,6 +163,14 @@ let
     ;
     pkgs = packages;
   };
+  meta = unwrapped.meta // {
+    position = ./. + (
+      if unwrapped.versionAttr.major == "3.10" then
+        "default"
+      else
+        unwrapped.versionAttr.major
+      ) + ".nix:0";
+  };
   self = if doWrap then
     stdenv.mkDerivation {
       inherit pname version passthru;
@@ -197,11 +205,12 @@ let
           wrapProgram "$i" ${makeWrapperArgs}
         done
       '';
-      inherit (unwrapped) meta;
+      inherit meta;
     }
   else
     unwrapped.overrideAttrs(_: {
       inherit passthru;
+      inherit meta;
     })
   ;
 in self
```

With and without this patch, this command prints the same output:

```sh
nix eval --json -L --impure --expr '
  let
    pkgs = (builtins.getFlake "'$PWD'").legacyPackages.${builtins.currentSystem}; 
  in
    pkgs.gnuradio.meta.position
  ' | jq --raw-output '.'
```

Output:

```
/nix/store/gzkvmymnwsd2r3ypzyxfpn7w4vsmjxqq-source/pkgs/applications/radio/gnuradio/shared.nix:127
```

Instead of the expected:

```
/nix/store/gzkvmymnwsd2r3ypzyxfpn7w4vsmjxqq-source/pkgs/applications/radio/gnuradio/default.nix:0

**Comment by Mic92** [comment]
Does lib.traceVal print your `meta.position` at all during eval?

**Comment by doronbehar** [comment]
If you meant this:

```
nix eval --json -L --impure --expr '
  let
    pkgs = (builtins.getFlake "'$PWD'").legacyPackages.${builtins.currentSystem}; 
  in
    pkgs.lib.traceVal pkgs.gnuradio.meta.position
  ' | jq --raw-output '.'
```

Then the output is:

```
trace: /nix/store/ram9z8s3vawgf6hkf5wm5q2jlwlyi401-source/pkgs/applications/radio/gnuradio/shared.nix:127
/nix/store/ram9z8s3vawgf6hkf5wm5q2jlwlyi401-source/pkgs/applications/radio/gnuradio/shared.nix:127
```

**Comment by doronbehar** [comment]
Even for the simplest package:

```diff
diff --git i/pkgs/by-name/he/hello/package.nix w/pkgs/by-name/he/hello/package.nix
index 7b8db4c7c3c2..c96f372399d8 100644
--- i/pkgs/by-name/he/hello/package.nix
+++ w/pkgs/by-name/he/hello/package.nix
@@ -52,6 +52,7 @@ stdenv.mkDerivation (finalAttrs: {
     license = licenses.gpl3Plus;
     maintainers = [ maintainers.eelco ];
     mainProgram = "hello";
+    position = "bla.nix";
     platforms = platforms.all;
   };
 })
```

This command:

```sh
nix eval --json -L --impure --expr '
  let
    pkgs = (builtins.getFlake "'$PWD'").legacyPackages.${builtins.currentSystem}; 
  in
    pkgs.lib.traceVal pkgs.hello.meta.position
  ' | jq --raw-output '.'
```

Prints:

```
trace: /nix/store/g6inivzr0ji1ii58b3qyxaxsaz8my5xh-source/pkgs/by-name/he/hello/package.nix:45
/nix/store/g6inivzr0ji1ii58b3qyxaxsaz8my5xh-source/pkgs/by-name/he/hello/package.nix:45
```

pbsds:
we could move the `meta.position` location to have the lowest priority

doronbehar:
> we could move the `meta.position` location to have the lowest priority

That would work for me probably. I hope though that it won't break your setup with the `mpvScripts` - a setup which I don't understand how it used the `meta.position` attribute - I Read a bit the related Nix files, like `buildLua.nix`, and still didn't manage to figure this out myself - a few examples like the above with `nix eval` commands would be great.

pbsds:
It seems as of March that mkDerivation overrides meta.position:

https://github.com/NixOS/nixpkgs/blob/5c1fa02dfca6c20c567159bbda9c0e7a0ac66b53/pkgs/stdenv/generic/make-derivation.nix#L522-L527

https://github.com/NixOS/nixpkgs/blob/5c1fa02dfca6c20c567159bbda9c0e7a0ac66b53/pkgs/stdenv/generic/make-derivation.nix#L564

https://github.com/NixOS/nixpkgs/blob/5c1fa02dfca6c20c567159bbda9c0e7a0ac66b53/pkgs/stdenv/generic/check-meta.nix#L486-L488

Proposed fix: https://github.com/NixOS/nixpkgs/pull/329900

nix_update/eval.py
```diff
@@ -142,13 +142,22 @@ def eval_expression(
...
-  position = if pkg ? isRubyGem then
+  position = if pkg ? meta.position then
+    sanitizePosition (positionFromMeta pkg)
+  else if pkg ? isRubyGem then
     raw_version_position
   else if pkg ? isPhpExtension then
     raw_version_position
-   else
+  else
     sanitizePosition (builtins.unsafeGetAttrPos "src" pkg);
```
doronbehar, line 161: Indeed if it is possible to set `meta.position`, this `nix-update` PR here makes a bit more sense. However, since `mkDerivation` always sets `meta.position`, the `pkg.src` attribute will never be retrieved here. Or perhaps it will be retrieved for other derivation builders? I doubt if there are builders that are not based on `mkDerivation` for which it will happen.

pbsds:
I revert back to my initial proposal in https://github.com/Mic92/nix-update/pull/267, please test it out and give us feedback :)

PR #251, opened 2024-05-21
use `nixfmt` with `--format`

getchoo:
following [rfc 166](https://github.com/NixOS/rfcs/pull/166), [nixfmt](https://github.com/NixOS/nixfmt) is now the official formatter for nix and slowing becoming more commonplace in both nixpkgs and the community. though it isn't stable just yet, i think it might be good to start using it sooner rather than later

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -21,7 +21,7 @@ pkgs.python311.pkgs.buildPythonApplication {
   makeWrapperArgs = [
     "--prefix PATH"
     ":"
-    (pkgs.lib.makeBinPath [ pkgs.nixVersions.stable or pkgs.nix_2_4 pkgs.nixpkgs-review pkgs.nix-prefetch-git pkgs.nixfmt-rfc-style ])
+    (pkgs.lib.makeBinPath [ pkgs.nixVersions.stable or pkgs.nix_2_4 pkgs.nixpkgs-review pkgs.nix-prefetch-git ])
   ];
   shellHook = ''
     # workaround because `python setup.py develop` breaks for me
--- a/default.nix
+++ b/default.nix
@@ -21,7 +21,7 @@ pkgs.python311.pkgs.buildPythonApplication {
   makeWrapperArgs = [
     "--prefix PATH"
     ":"
-    (pkgs.lib.makeBinPath [ pkgs.nixVersions.stable or pkgs.nix_2_4 pkgs.nixpkgs-review pkgs.nix-prefetch-git ])
+    (pkgs.lib.makeBinPath [ pkgs.nixVersions.stable or pkgs.nix_2_4 pkgs.nixpkgs-review pkgs.nix-prefetch-git pkgs.nixfmt-rfc-style ])
   ];
   shellHook = ''
     # workaround because `python setup.py develop` breaks for me
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ tested. Additionally, the `--review` flag can be used to
 initiate a run of [nixpkgs-review](https://github.com/Mic92/nixpkgs-review), which will ensure all
 dependent packages can be built. In order to ensure consistent
 formatting, the `--format` flag will invoke
-[nixpkgs-fmt](https://github.com/nix-community/nixpkgs-fmt).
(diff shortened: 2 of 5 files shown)


SuperSandro2000:
IMO we should probably drop that option entirely from a tool that is about updating packages since those changes also most of the time should get their own commit.

getchoo:
i personally use this a lot to make sure that any changes i've (or nix-update) have made are still inline with the previous styling. i've been doing this less lately mainly because it's still using nixpkgs-fmt

i think it may be nice to have an option for a separate format commit, though - as that would be more inline with what we should be doing

default.nix
```diff
@@ -21,7 +21,7 @@ pkgs.python311.pkgs.buildPythonApplication {
   makeWrapperArgs = [
     "--prefix PATH"
     ":"
-    (pkgs.lib.makeBinPath [ pkgs.nixVersions.stable or pkgs.nix_2_4 pkgs.nixpkgs-review pkgs.nix-prefetch-git ])
+    (pkgs.lib.makeBinPath [ pkgs.nixVersions.stable or pkgs.nix_2_4 pkgs.nixpkgs-review pkgs.nix-prefetch-git pkgs.nixfmt-rfc-style ])
```
Mic92, line 24: I would like to avoid haskell here because it doesn't work on some platforms
and adds quite a bit of dependencies:

```suggestion
    (pkgs.lib.makeBinPath [ pkgs.nixVersions.stable or pkgs.nix_2_4 pkgs.nixpkgs-review pkgs.nix-prefetch-git ])
```

This is also because `--format` is likely a rare use case and most people shouldn't need to download it.

Mic92:
@mergify queue

Mic92:
> IMO we should probably drop that option entirely from a tool that is about updating packages since those changes also most of the time should get their own commit.

I think that's too late now. I feel like already too many people depend on it: https://github.com/Mic92/nix-update/issues/253

Mic92:
@mergify rebase

colemickens:
```
  $ nixfmt /var/lib/github-actions-runners/raisin-default/work/nixcfg/nixcfg/pkgs/alacritty/default.nix
    File "/nix/store/kjknpmby5d3ff565aq7ldmklzr2pxfx8-nix-update-1.0.0/bin/.nix-update-wrapped", line 9, in <module>
      sys.exit(main())
               ^^^^^^
    File "/nix/store/kjknpmby5d3ff565aq7ldmklzr2pxfx8-nix-update-1.0.0/lib/python3.11/site-packages/nix_update/__init__.py", line 329, in main
      run(["nixfmt", package.filename], stdout=None)
    File "/nix/store/kjknpmby5d3ff565aq7ldmklzr2pxfx8-nix-update-1.0.0/lib/python3.11/site-packages/nix_update/utils.py", line 37, in run
      return subprocess.run(
             ^^^^^^^^^^^^^^^
    File "/nix/store/6b1fqdwb3g56j5pazv8zkx9qd0mv3wiz-python3-3.11.9/lib/python3.11/subprocess.py", line 548, in run
      with Popen(*popenargs, **kwargs) as process:
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/nix/store/6b1fqdwb3g56j5pazv8zkx9qd0mv3wiz-python3-3.11.9/lib/python3.11/subprocess.py", line 1026, in __init__
      self._execute_child(args, executable, preexec_fn, close_fds,
    File "/nix/store/6b1fqdwb3g56j5pazv8zkx9qd0mv3wiz-python3-3.11.9/lib/python3.11/subprocess.py", line 1955, in _execute_child
      raise child_exception_type(errno_num, err_msg, err_filename)
  FileNotFoundError: [Errno 2] No such file or directory: 'nixfmt'
  ```

I get this recently, seems related to this PR.

Guanran928:
> ```
>   FileNotFoundError: [Errno 2] No such file or directory: 'nixfmt'
> ```

You might need nixfmt in your $PATH

colemickens:
sure, `nixfmt` docs say its in nixpkgs, but it's not. So I guess I'll go hunt the derivation name.

I guess it's `nixfmt-rfc-style`. Thanks.

PR #254, opened 2024-06-01
Ignore yanked versions in `fetchCrate`

marsam:


Mic92:
Thanks a lot!

Mic92:
@mergify queue

Issue #257, opened 2024-06-09
Support Gitea instances other than codeberg.org or gitea.com


pluiedev:
These two are the only sites that would be recognized as Gitea in https://github.com/Mic92/nix-update/blob/master/nix_update/version/gitea.py#L10, despite the rising popularity of self-hosting Gitea/Forgejo/... instances. It would be nice to test if the given URL is hosted on Gitea or not, using an API check, or at least give us the option to force nix-update to treat the URL as a link to a Gitea repository.

marsam:
I think it is doable.

I created a quick wip https://github.com/marsam/nix-update/commit/565961d
@Mic92 please let me know if you think it could be an acceptable PR

Mic92:
Sure. My only worry is that the version api might collide with other software.

Hopefully those will never become valid in future:

https://crates.io/api/v1/version
https://registry.npmjs.org/api/v1/version

There is also https://git.clan.lol/api/v1/signing-key.gpg, where we could check for a HTTP 200 code, which seems like a bit less generic api.

SuperSandro2000:
I look at the swagger and I didn't find a generic endpoint where you could extract something gitea specific.

pluiedev:
Maybe then we should allow people to specifically try using a certain fetcher? (e.g. the `--fetcher` option in nurl) We could also maybe try to detect which fetcher is used in the Nix expression, so if it sees `src = fetchFromGitea` it would try to use Gitea

Mic92:
Maybe fetchFromGitea can inject some meta attribute that allows us to detect gitea?

Mic92:
Something like this?

```
diff --git a/pkgs/build-support/fetchgitea/default.nix b/pkgs/build-support/fetchgitea/default.nix
index 513ceba861cb..ceb29c59433c 100644
--- a/pkgs/build-support/fetchgitea/default.nix
+++ b/pkgs/build-support/fetchgitea/default.nix
@@ -5,5 +5,7 @@
 lib.makeOverridable (
 { domain, ... }@args:

-fetchFromGitHub ((removeAttrs args [ "domain" ]) // { githubBase = domain; })
-)
+fetchFromGitHub ((removeAttrs args [ "domain" ]) // {
+  githubBase = domain;
+  meta.gitea = true;
+}))
```

PR #259, opened 2024-06-11
Add check whether a host is a Gitea instance to fetcher

marsam:
To allow self-hosted Gitea instances

Closes https://github.com/Mic92/nix-update/issues/257


Mic92:
@mergify queue

PR #260, opened 2024-06-25
update Cargo.lock after other hashes

getchoo:
As `update_cargo_lock()` overrides the top level derivations, this will cause hash mismatch errors in the event of other FODs being used in the environment (i.e., `pnpmDeps`) during updates if they also need to be updated. `gitbutler` in nixpkgs at [this](https://github.com/NixOS/nixpkgs/tree/e6e4cd92ad886b91a6de120ce61c81b7e6072530/pkgs/by-name/gi/gitbutler) revision is a real world example of this occurring, and changing the order of these function calls allows it to run as expected


Mic92:
@mergify queue

PR #261, opened 2024-07-04
README: Clarify that nix-update *does* update Cargo.lock files

Pandapip1:
See https://github.com/NixOS/nixpkgs/pull/324401#discussion_r1664866850


Mic92:
@mergify queue

PR #264, opened 2024-07-14
Add support for updating Mix dependencies for Elixir projects

mattpolzin:
Happy to move things around, rename, document differently. This adds support for automatically updating the `mixFodDeps` hash used by `beamPackages.mixRelease`.

Thanks for the handy update script!


Mic92:
Could you add some test please like we have for the other implementations? Otherwise it looks good to me.

mattpolzin:
Oh yeah absolutely. I thought to add tests and then lost track of that. Will do!

mattpolzin:
@Mic92 there's no hurry here from my perspective, just wanted to mention that I added a test.

Mic92:
@mergify queue

Issue #164, opened 2023-05-18
Unable to generate a `Cargo.lock` file for repositories that do not have one


nickgerace:
It would be great if `nix-update` had the ability to generate `Cargo.lock` file. Of course, this would have to be explicit and not default behavior.

We encountered this issue when working on adding [reindeer](https://github.com/facebookincubator/reindeer) to nixpkgs. More context here: https://github.com/NixOS/nixpkgs/pull/232699#discussion_r1198358248.

doronbehar:
Similarly, it'd be great if there was also support for updating vendored `package-lock.json` files.

nickgerace:
Thank you @natsukium @Mic92!

PR #266, opened 2024-08-09
Add support for generating Cargo.lock

natsukium:
closes https://github.com/Mic92/nix-update/issues/164

The problems with vendored Cargo.lock to nixpkgs are discussed in the following [issue](https://github.com/NixOS/nixpkgs/issues/327063), but for now it would be a good idea to allow such Cargo.lock to be updated with nix-update.
There are many projects, especially in python packages, that have no Cargo.lock, and we feel painful updating them for a while until the issue is resolved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ designed to work with nixpkgs but also other package sets.
 - update mixRelease's mixFodDeps
 - update fetchYarnDeps offlineCache output hash
 - update flake outputs (see `--flake`)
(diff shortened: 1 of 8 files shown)


Mic92:
nix-update also works outside of nixpkgs, where lock size is less of an issue.

Mic92:
@mergify rebase

Mic92:
@mergify queue

PR #267, opened 2024-08-10
Try pkg.meta.position after unsafeGetAttrPos

pbsds:
`meta.position` is prone to point at files where `src` and `version` is not contained, as is the case for overridden packages and packages that use a generic builder.
There is also the fix explored in https://github.com/NixOS/nixpkgs/pull/329900 and https://github.com/NixOS/nixpkgs/pull/295973, which would alleviate many of the errors while creating new ones.

Fixes https://github.com/Mic92/nix-update/pull/247#issuecomment-2281135694

### DIFF
--- a/nix_update/eval.py
+++ b/nix_update/eval.py
@@ -153,14 +153,14 @@ let
 
   raw_version_position = sanitizePosition (builtins.unsafeGetAttrPos "version" pkg);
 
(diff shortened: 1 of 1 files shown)


Mic92:
@mergify queue

doronbehar:
Fixes the issue for me :) Thanks.

PR #268, opened 2024-08-11
make updating cargo.lock more robust

Mic92:
* don't assume we have a nix flake registry instead take cargo from the build
* set experimental flags
* apply patches to build


Mic92:
@mergify queue

PR #273, opened 2024-08-21
fix handling of cargo.lock outside a workspace

natsukium:
This is a follow-up PR for https://github.com/Mic92/nix-update/pull/266.

If Cargo.toml is outside a workspace, Cargo.lock is generated in the same directory as Cargo.toml.

The current implementation expects Cargo.lock to be in the project root, so the update fails.
`FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmpn_dgfq0d/Cargo.lock'`


Mic92:
@mergify queue

PR #269, opened 2024-08-15
github: Add fallback to atom feed if project does not use releases

jvanbruegge:
As requested in #246

PR #274, opened 2024-09-06
github: Revert use of API to determine latest version

### LINKED ISSUE #272: Github releases may not be the latest
Hi,

some projects may release new minor/patch releases after a new major/minor.

For example, at the time of writing https://github.com/kubernetes/kubernetes/tags:

* v1.30.4
* v1.29.8
* v1.28.13
* v1.31.0
 
therefore:

```console
$ nix --extra-experimental-features "flakes nix-command" run github:Mic92/nix-update -- --flake kubernetes
...
trying to fetch https://api.github.com/repos/kubernetes/kubernetes/releases
Not updating version, already 1.30.4
```

`nix-update` assumes that `1.30.4` is the latest but it is not.

Any suggestions?

Thanks

lilyinstarlight:
The GitHub API behavior does not work for any repository that has used GitHub releases in the past but only uses tags for newer versions. The original motivation for the change was to prevent pre-releases from being selected as the latest version, but the `--version-regex` flag exists precisely for the purpose of selecting only intended versions (and not unrelated versions, pre-releases, dev tags, etc)

As such, the change provided no benefit to the package listed in the motivation, but it did break updates for quite a few packages and has resulted in downgrades and misversioning, among other things

A possible attempt to include the GitHub API could be used to specifically exclude tags associated with pre-releases in the GitHub API, but otherwise continue using tags as the "source of truth" for the latest version. This change does not attempt that as it is only intended to make nix-update work correctly as it did before the change. Feel free to ping me if you attempt to do that as a way forward @jvanbruegge :)

This reverts commits bf33dc45ce26f36233fd58790f90dc8710ca7198 and 696efb62a9df3223a255b15e005e4d0815130a0c

Reference #246 (also note <https://github.com/Mic92/nix-update/pull/246#issuecomment-2323018854>) #269
Fixes #272
Notify @jvanbruegge @Mic92


jian-lin:
nix-update in Nixpkgs does not have this patch, so it keeps opening "wrong" PRs such as https://github.com/NixOS/nixpkgs/pull/342493.

What about cutting a new release or adding this patch to Nixpkgs?

Mic92:
release added.

PR #279, opened 2024-09-13
Add --update-script-args

pbsds:
This allows you to skip the prompt when using `-u`, nice to have when scripting `nix-update`

```
nix-update sbt --use-update-script --update-script-args "--argstr skip-prompt true"
```


Mic92:
@mergify queue

PR #296, opened 2024-11-10
Fix updating all matching version numbers

ayyess:
Prior to this change, nix-update would update all version strings which were equal to the specified package to update. Now, only the version line of the specified package will be updated.

### DIFF
--- a/nix_update/update.py
+++ b/nix_update/update.py
@@ -35,11 +35,23 @@ def replace_version(package: Package) -> bool:
 
     if changed:
         info(f"Update {old_version} -> {new_version} in {package.filename}")
+        version_string_in_version_declaration = False
+        if package.version_position is not None:
+            with open(package.filename) as f:
+                for i, line in enumerate(f, 1):
+                    if package.version_position.line == i:
+                        version_string_in_version_declaration = old_version in line
+                        break
         with fileinput.FileInput(package.filename, inplace=True) as f:
-            for line in f:
+            for i, line in enumerate(f, 1):
                 if package.new_version.rev:
                     line = line.replace(package.rev, package.new_version.rev)
-                print(line.replace(f'"{old_version}"', f'"{new_version}"'), end="")
+                if (
+                    not version_string_in_version_declaration
+                    or package.version_position.line == i
+                ):
+                    line = line.replace(f'"{old_version}"', f'"{new_version}"')
+                print(line, end="")
     else:
         info(f"Not updating version, already {old_version}")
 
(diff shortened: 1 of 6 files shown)


tests/testpkgs/default.nix
```diff
@@ -28,4 +28,5 @@
   pnpm = pkgs.callPackage ./pnpm.nix { };
   maven = pkgs.callPackage ./maven.nix { };
   mix = pkgs.callPackage ./mix.nix { };
+  set = import ./set.nix { inherit pkgs; };
```
SuperSandro2000, line 31: This should use callPackage, too.

tests/testpkgs/set.nix
```diff
@@ -0,0 +1,26 @@
...
+      rev = "v${version}";
+      sha256 = "sha256-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=";
+    };
+  };
+
+  skim = pkgs.stdenv.mkDerivation rec {
+    pname = "skim";
+    version = "0.0.0";
+
+    src = pkgs.fetchFromGitHub {
```
SuperSandro2000, line 19:
```suggestion
{ stdenv, fetchFromGitHub }:
{
  fd = stdenv.mkDerivation rec {
    pname = "fd";
    version = "0.0.0";

    src = fetchFromGitHub {
      owner = "sharkdp";
      repo = pname;
      rev = "v${version}";
      sha256 = "sha256-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=";
    };
  };

  skim = stdenv.mkDerivation rec {
    pname = "skim";
    version = "0.0.0";

    src = fetchFromGitHub {
```

ayyess:
I'll look at fixing the treefmt tests too.

nix_update/update.py
```diff
@@ -36,10 +36,15 @@ def replace_version(package: Package) -> bool:
     if changed:
         info(f"Update {old_version} -> {new_version} in {package.filename}")
         with fileinput.FileInput(package.filename, inplace=True) as f:
-            for line in f:
+            for i, line in enumerate(f, 1):
                 if package.new_version.rev:
                     line = line.replace(package.rev, package.new_version.rev)
-                print(line.replace(f'"{old_version}"', f'"{new_version}"'), end="")
+                if (
+                    package.version_position is None
```
Mic92, line 43: Mhm. It's nice that we are more precise but is this not going to be a problem for the a lot of case i.e. if a user does this:

```
let
  version = "1.0.0";
in stdenv.mkDerivation {
  name = "package-${version}";
}
```

I see these cases actually more often than multiple packages in a file.

Mic92: Maybe we should do the following. First an iteration where we check if the version line has our version and if this doesn't work it will backtrack and also replace other assignments.

ayyess: Good point. I've added the precheck iteration but this handling isn't as simple

Mic92:
@mergify queue

PR #300, opened 2024-11-18
feat: add support for fetchCargoVendor

### LINKED ISSUE #298: Add support for fetchCargoVendor-based cargoDeps
Now that https://github.com/NixOS/nixpkgs/pull/349360 is merged we should probably add support for updating packages using it.

This does not work with the old method because `cargoDeps.outputHash` does not exist.
Instead `cargoDeps.vendorStaging.outputHash` should be used.

GaetanLepage:
Add support for updating packages using the newly introduce [`fetchCargoVendor`](https://github.com/NixOS/nixpkgs/pull/349360).

Fixes #298

### DIFF
--- a/nix_update/eval.py
+++ b/nix_update/eval.py
@@ -53,6 +53,7 @@ class Package:
     go_modules: str | None
     go_modules_old: str | None
     cargo_deps: str | None
+    cargo_vendor_deps: str | None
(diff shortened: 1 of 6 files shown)


Mic92:
@mergify queue

PR #301, opened 2024-11-18
feat: Fix --override-filename for flakes

### LINKED ISSUE #282: git diff call fails, ".../default.nix" is outside repository at "/repo/path"
```
nix-update \
--flake \
packages.x86_64-linux.foot \
--version \
branch=master \
--commit \
--override-filename \
./pkgs/foot/metadata.nix
```

results in:

```
[the nix expr that nix-update evals + prints is omitted, except the last line of the output here:]
$ git diff --color=never --word-diff=porcelain -- /nix/store/hhwdvcdgz7qz2prms49lr1h82v5qagxp-source/pkgs/foot/default.nix
fatal: /nix/store/hhwdvcdgz7qz2prms49lr1h82v5qagxp-source/pkgs/foot/default.nix: '/nix/store/hhwdvcdgz7qz2prms49lr1h82v5qagxp-source/pkgs/foot/default.nix' is outside repository at '/home/cole/code/nixpkgs-wayland
[traceback omitted]
```

(note this is not with master of `nixpkgs-wayland`, but a branch where I'm experimenting with `nix-update` instead of my old bespoke update script)

Not quite sure what's going on here? I'm passing a relative path for the file for it to try and update, I'm not doing anything to make an out of repo reference, as far as I can tell.

ede1998:
Closes #282

I ran `pytest` and tested it for a package in a flake.
I didn't test without flake as I don't know how to do that atm.

I'm a bit unsure with `self.version_position.file = self.filename` as I'm overwriting the filename unconditionally but if it wasn't passed by `--override-filename` but comes from the nix expression instead, it might not have the same source as the `raw_version_positon` (`sanitizePosition (builtins.unsafeGetAttrPos "version" pkg)` vs potentially `sanitizePosition (positionFromMeta pkg)`  or `sanitizePosition (builtins.unsafeGetAttrPos "src" pkg)` ), which could break some other cases.


Mic92:
You can run the formatter by using the `nix fmt` command.

ede1998:
Thank you. I ended up running `nixfmt` directly.

I was just confused with the `treefmt` stuff.

Mic92:
@mergify queue

PR #305, opened 2024-12-12
fix: use `shlex.join` for command formatting in run function

trim21:
current command output is not very nice when some argument contains space, for example: `$ git -C /export/nvme-512g/src/nixpkgs commit --verbose --message rtorrent: 0.10.0-unstable-2024-12-06 -> 0.10.0-unstable-2024-12-08 /export/nvme-512g/src/nixpkgs/pkgs/by-name/rt/rtorrent/package.nix`

### DIFF
--- a/nix_update/utils.py
+++ b/nix_update/utils.py
@@ -1,4 +1,5 @@
 import os
+import shlex
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


Mic92:
@mergify queue

PR #309, opened 2024-12-16
feat: add support for generation package-lock.json

Moraxyc:
- **feat: add support for generating package-lock.json**
- **refactor: unify lockfile generation functions**

Combine cargo and npm lockfile generation into a single function `generate_lockfile` use parameters to handle different types and configurations reduce code duplication and improve maintainability.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,8 +25,8 @@ work with nixpkgs but also other package sets.
 - update mixRelease's mixFodDeps
 - update fetchYarnDeps offlineCache output hash
 - update flake outputs (see `--flake`)
-- generate the following lockfile, Cargo.lock (see `--generate-lockfile` and
-  `--lockfile-metadata-path`)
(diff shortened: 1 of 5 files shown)


Moraxyc:
We should also consider the possibility that the package uses `npm_deps_hash`,` cargo_lock`, and `--generate-lockfile` together. IMO, there are two ways to address this:

Separate Flags: Introduce additional flags, such as `--generate-npm-lockfile`. To maintain backward compatibility, it’s best to retain the existing `--generate-lockfile` flag. However, the downside is that if `nix-update` supports more lockfile types in the future, managing multiple flags could become messy.

Parameterized `--generate-lockfile`: Allow `--generate-lockfile` to accept parameters. By default, if no parameter is provided, it generates `cargoLock`. If parameters are specified, they can be listed as a comma-separated string (e.g., "cargo,npm") to indicate which lockfiles to generate.

Mic92:
@mergify queue

PR #311, opened 2024-12-17
feat: add `--subpackage` option to update multiple derivations

### LINKED ISSUE #310: feature request: update multiple derivations in one package
Nix-update currently doesn't have any way of knowing when a package has multiple derivations that need to be updated. I ran into this when packaging autobrr in [nixos/nixpkgs#287593](https://github.com/NixOS/nixpkgs/pull/287593): autobrr consists of a frontend built with pnpm, which is incorporated into the application's binary when the backend is built as a Go module.

I don't feel like writing my own update script, and I don't think breaking the package up into two files makes sense here (since neither part can work independently of the other). To fix this issue in general, I believe that nix-update would need to support multiple derivations in one package, both of which have hashes that need to be updated.

av-gal:
Adds an option for users to specify subpackages that `nix-update` should also look at. `nix-update` will evaluate these sub-derivations after updating the package's version and source, but before trying to evaluate the package's main derivation. This allows for subpackages that are depended upon by the main package build.

Not included is any sort of accommodation for sub-subpackages. I hope those do not occur often in real builds.

Closes #310.

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,6 +197,63 @@ Arguments can be passed to `nix-shell maintainers/scripts/update.nix` like so
 $ nix-update sbt --use-update-script --update-script-args "--argstr skip-prompt true"
 ```
 
+## Subpackages
+
+Some packages consist of multiple fixed-output derivations derived from the same
+upstream source. For example, a Go project with Go module dependencies might
+also include a JavaScript project with npm dependencies.
+
+To support such use cases, `nix-update` allows specifying subpackages directly
+in the command line. Consider a package accessible via the `some-package`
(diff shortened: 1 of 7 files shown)


Mic92:
Can we get a test for this?

av-gal:
Test added, it's just a simplified version of what I'm packaging. Thanks for your time, please let me know if there's anything else.

Mic92:
@av-gal I added some documentation. Please check.

av-gal:
I fixed a typo in the name of the option, and expanded your comment about repeating the option for multiple subpackages. Otherwise, LGTM.

Mic92:
@mergify queue

Issue #304, opened 2024-12-11
crash with fetchFromGitHub tag


gepbird:
## Describe the bug

<!-- A clear and concise description of what the bug is. -->

When using `nix-update` in an update script, and the `src` being `fetchFromGitHub` with a `tag` attribute rather than `rev`, the script crashes.

## Steps To Reproduce

*This is not a minimal reproducible example but based on a PR when I found this issue*


1. clone/checkout https://github.com/gepbird/nixpkgs/tree/nix-update-tag-crash-demo
2. `nix-shell maintainers/scripts/update.nix --argstr package ps3-disc-dumper`
3. observe the nix-update crash

<details><summary>logs</summary>

```
nixpkgs ❯ nix-shell maintainers/scripts/update.nix --argstr package ps3-disc-dumper                                                                                                                               
                                                                                                                                                                                                                  
                                                                                                                                                                                                                  
Going to be running update for following packages:                                                                                                                                                                
 - ps3-disc-dumper-4.2.5                                                                                                                                                                                          
                                                                                                                                                                                                                  
Press Enter key to continue...                                                                                                                                                                                    
Running update for:                                                                                                                                                                                               
 - ps3-disc-dumper-4.2.5: UPDATING ...                                                                                                                                                                            
 - ps3-disc-dumper-4.2.5: ERROR                                                                                                                                                                                   
                                                                                                                                                                                                                  
--- SHOWING ERROR LOG FOR ps3-disc-dumper-4.2.5 ----------------------                                   
                                                                                                                                                              
error:                                                                                                   
       … while evaluating attribute 'hash'                                                               
         at «string»:37:3:              
           36|   rev = pkg.src.rev or null;                                                                                                                                                                                                  
           37|   hash = pkg.src.outputHash or null;
             |   ^                                  
           38|   go_modules = pkg.goModules.outputHash or null;                                                                                                                                                                              
                                                                                                         
       … while evaluating the attribute 'src.outputHash'                                                                                                      
         at /home/gep/nixpkgs/pkgs/by-name/ps/ps3-disc-dumper/package.nix:14:3:                                                                                                                                                              
           13|                                                                                           
           14|   src = fetchFromGitHub {          
             |   ^                                                                                                                                                                                                                           
           15|     owner = "13xforever";                                                                                                                      
                                                                                                         
       (stack trace truncated; use '--show-trace' to show the full, detailed trace)                                                                                                                                                          
                                                                                                                                                                                                                  
[...]

PR #312, opened 2024-12-23
feat: tag in fetchFromGitHub support

gepbird:
Mainly opening this to avoid duplicated work as I see more people want this feature.

I don't have much time right now, but I can probably finish it in a day.

Closes https://github.com/Mic92/nix-update/issues/304, https://github.com/NixOS/nixpkgs/issues/367562

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ $ nix run github:Mic92/nix-update
 ## USAGE
 
 First change to your directory containing the nix expression (Could be a nixpkgs
-or your own repository). Than run `nix-update` as follows
+or your own repository). Then run `nix-update` as follows
 
 ```console
 $ nix-update attribute [--version version]
@@ -77,7 +77,7 @@ If your package is defined in a flake use the `--flake` flag instead:
 $ nix-update attribute --flake [--version version]
 ```
 
-`nix-update` will than try to update either the
+`nix-update` will then try to update either the
 `packages.{currentSystem}.{attribute}` or `{attribute}` output attribute of the
 given flake. To update a package in `legacyPackages`, pass the full path to that
 package including the platform: `legacyPackages.{platform}.{attribute}`.
@@ -260,10 +260,10 @@ First clone the repo to your preferred location (in the following, we assume
 `~/` - your home):
 
 ```console
-$ git clone https://github.com/Mic92/nix-update/ ~/nix-update
+$ git clone https://github.com/Mic92/nix-update/ ~/git/nix-update
 ```
 
-Than enter the dev shell:
+Then enter the dev shell:
 
 ```console
 $ cd ~/nix-update
@@ -297,7 +297,7 @@ $ nix fmt
 
 ## Known Bugs
 
-nix-update might not work correctly if a file contain multiple packages as it
+nix-update might not work correctly if a file contains multiple packages as it
 performs naive search and replace to update version numbers. This might be a
 problem if:
 
--- a/flake.lock
+++ b/flake.lock
@@ -7,11 +7,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1730504689,
-        "narHash": "sha256-hgmguH29K2fvs9szpq2r3pz2/8cJd2LPS+b4tfNFCwE=",
+        "lastModified": 1733312601,
+        "narHash": "sha256-4pDvzqnegAfRkPwO3wmwBhVi/Sye1mzps0zHWYnP88c=",
         "owner": "hercules-ci",
         "repo": "flake-parts",
-        "rev": "506278e768c2a08bec68eb62932193e341f55c90",
+        "rev": "205b12d8b7cd4802fbcb8e8ef6a0f1408781a4f9",
         "type": "github"
       },
       "original": {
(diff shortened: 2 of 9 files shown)


Mic92:
Looks good. Take your time. I won't get to it, but I will at least timely merge it.

gepbird:
The test passes but there may be some things that the tests didn't catch (eg. https://github.com/Mic92/nix-update/pull/312/commits/3f6932b245c65d1021383382c61387398244c995), I will try to find these later.

The fetchers still use only rev (or only version_number sometimes?), adding tag support for them is out of the scope of this PR. This leaves the code base in a bit inconsistent state where sometimes we check `version.rev or version.tag` when the version came from the nix eval, but for versions coming from fetchers we still only do `new_version.rev`.

Another weird thing I found is: this might be already fixed in nixpkgs/master, I couldn't reproduce my issue.

gepbird:
I did some refactors to make it clearer when we are talking about a "rev or a tag" for the old/not-updated package and called it `old_rev_tag`. I think I caught all instances that previously checked for a `rev` (excluding versions coming from fetchers as explained above).

I think this is ready, but more manual testing wouldn't hurt :)

xiaoxiangmoe:
@gepbird Can you add both tag and rev support?

For example:

```nix
{
  buildGoModule,
  fetchFromGitHub,
  lib,
  nix-update-script,
}:
buildGoModule rec {
  pname = "lipo-go";
  version = "0.9.3";
  env = {
    GIT_VERSION = version;
    GIT_REVISION = "b7b34565565e3cde8037d1b5ee95dd2bb3579ef1";
  };
  src = fetchFromGitHub {
    owner = "konoui";
    repo = "lipo";
    tag = "v${version}";
    rev = env.GIT_REVISION;
    hash = "sha256-FW2mOsshpXCTTjijo0RFdsYX883P2cudyclRtvkCxa0=";
  };

  passthru.updateScript = nix-update-script { };
}
  ```

**Comment by gepbird** [comment]
@xiaoxiangmoe can you explain that use case? I don't see a reason for using both, because if they differ, which should be fetched?

And there's an eval failure on (a 4 day old) nixpkgs:
```
       error: fetchFromGitHub requires one of either `rev` or `tag` to be provided (not both).                                                                                                                    
```

**Comment by xiaoxiangmoe** [comment]
Some app build will need both GIT_VERSION and GIT_REVISION for log info, show version and so on. So we need get both tag and rev.

For example https://github.com/konoui/lipo

```
./result/bin/lipo -version                                       
0.9.2/b7b34565565e3cde8037d1b5ee95dd2bb3579ef1
```

**Comment by xiaoxiangmoe** [comment]
```
error: fetchFromGitHub requires one of either `rev` or `tag` to be provided (not both).    
```

I think we should support both for it. (Maybe someday tag will change, and rev is always immutable and relyable.) If tag and rev is not same, we can put a warning for users.

xiaoxiangmoe:
Here is another example
https://github.com/NixOS/nixpkgs/blob/a3484e1559a2e63670aef012b0afd1217a9e011b/pkgs/by-name/af/affine/package.nix#L46

I can't use nix-update-script because I can't update version with GITHUB_SHA both. And sentry relies on the env GITHUB_SHA.

gepbird:
> Some app build will need both GIT_VERSION and GIT_REVISION for log info, show version and so on. So we need get both tag and rev.
> 
> For example [konoui/lipo](https://github.com/konoui/lipo?rgh-link-date=2024-12-25T16%3A40%3A38Z)
> 
> ```
> ./result/bin/lipo -version                                       
> 0.9.2/b7b34565565e3cde8037d1b5ee95dd2bb3579ef1
> ```

In [this package](https://github.com/NixOS/nixpkgs/blob/a3484e1559a2e63670aef012b0afd1217a9e011b/pkgs/by-name/li/libnvidia-container/package.nix#L68-L69) some REV (COMMIT) and TAG is being patched from the `version` and `src.rev` attributes. Or [this go package](https://github.com/NixOS/nixpkgs/blob/cfeb7de067cb4fee49f7fef528e28d69c0dacb9a/pkgs/by-name/xq/xq-xml/package.nix#L25-L26) seems like a better example that overrides the commit (rev) and version (tag).

> ```
> error: fetchFromGitHub requires one of either `rev` or `tag` to be provided (not both).    
> ```
> 
> I think we should support both for it.
> 
> Maybe someday tag will change (there are precedents for this in history) and rev is always immutable and relyable.
> 
> If tag and rev is not same, we can put a warning for users.

I doubt it would change, any fetcher like `fetchFromGitHub` should only have one source, either a tag (that is just a named rev) or a rev. And validating whether the rev and the tag is the same would add extra overhead.

gepbird:
> Here is another example [NixOS/nixpkgs@`a3484e1`/pkgs/by-name/af/affine/package.nix#L46](https://github.com/NixOS/nixpkgs/blob/a3484e1559a2e63670aef012b0afd1217a9e011b/pkgs/by-name/af/affine/package.nix#L46)
> 
> I can't use nix-update-script because I can't update `version` with `GITHUB_SHA` both. And sentry relies on the env GITHUB_SHA.

This makes sense, but seems like a rare use case and implementing it wouldn't be easy, so it's out of the scope of this PR, sorry. After fetching the new version, we'd need to find the related hard coded commit SHA (this step would probably lead to a lot of false positives and breaking packages), then fetch the new commit SHA and replace the old one.

If you really need to update the commit SHA automatically, you can write an update script that utilises `nix-update` to update the version, src.hash and vendorHash, then you can probably obtain the commit SHA with GitHub's API and replace it with sed.

Feel free to open an issue on nixpkgs or start a discourse thread for help with packaging or making the update script and ping me :)

Mic92:
@mergify queue

PR #313, opened 2025-01-04
fix `fetchurl` with `--version-regex`

DimitarNestorov:
When using `--version-regex` on a package the fetches source via `fetchurl` the update failed because `package.rev` was `None`

### DIFF
--- a/nix_update/update.py
+++ b/nix_update/update.py
@@ -46,7 +46,7 @@ def replace_version(package: Package) -> bool:
                         break
         with fileinput.FileInput(package.filename, inplace=True) as f:
             for i, line in enumerate(f, 1):
-                if package.new_version.rev:
+                if package.rev is not None and package.new_version.rev:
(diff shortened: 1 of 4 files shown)


Mic92:
@mergify queue

Mic92:
Thanks!

nix_update/update.py
```diff
@@ -46,7 +46,7 @@ def replace_version(package: Package) -> bool:
                         break
         with fileinput.FileInput(package.filename, inplace=True) as f:
             for i, line in enumerate(f, 1):
-                if package.new_version.rev:
+                if package.ref is not None and package.new_version.rev:
```
Mic92, line 49:
```suggestion
                if package.rev is not None and package.new_version.rev:
```

Mic92:
@mergify queue

Mic92:
@mergify refresh

Mic92:
@mergify dequeue

PR #324, opened 2025-02-15
feat: add --src-only flag

gepbird:
Context: last few comments in https://github.com/Mic92/nix-update/pull/320 (this flag was initially mentioned as `--skip-deps`,  `--dont-update-deps` or `--no-fetch-deps` that would've only applied to `nugetDeps`)

In some cases you want to use nix-update's logic where it can fetch the latest version and hash from various sources, but don't want to update dependencies like `nugetDeps`, `npmDeps`, `cargoDeps`, etc. Currently the solution for this in nixpkgs is making a request to the GitHub API or other sources, and using `update-source-version` to update the version, then manually do the deps update part.

Instead of writing 
```bash
new_version="$(curl -s "https://api.github.com/repos/AvaloniaUI/Avalonia/releases/latest" | jq --raw-output '.tag_name')"
update-source-version avalonia "$new_version"
# operations that need to be ran before updating deps would go here
"$(nix-build -A "$package".fetch-deps --no-out-link)"
```
You can do
```bash
nix-update avalonia --src-only
# operations that need to be ran before updating deps would go here
"$(nix-build -A "$package".fetch-deps --no-out-link)"
```

cc @Mic92 @corngood @GGG-Killer

### DIFF
--- a/nix_update/__init__.py
+++ b/nix_update/__init__.py
@@ -116,6 +116,11 @@ def parse_args(args: list[str]) -> Options:
         default=None,
     )
     parser.add_argument(
+        "--src-only",
+        help="Only update the source, not dependencies such as npmDeps, cargoDeps or nugetDeps",
+        action="store_true",
+    )
+    parser.add_argument(
         "--option",
         help="Nix option to set",
         action="append",
@@ -154,6 +159,7 @@ def parse_args(args: list[str]) -> Options:
         system=a.system,
         generate_lockfile=a.generate_lockfile,
         lockfile_metadata_path=a.lockfile_metadata_path,
+        src_only=a.src_only,
         extra_flags=extra_flags,
     )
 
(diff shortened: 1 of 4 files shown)


GGG-KILLER:
Minor nit (and the start of bikeshedding, I'm sorry): I'd call it `--src-only` since not all builders use `deps` in the name.

gepbird:
> Minor nit (and the start of bikeshedding, I'm sorry): I'd call it `--src-only` since not all builders use `deps` in the name.

That sounds more correct. It would take much more effort to change it later, so thanks for suggesting it early!

GGG-KILLER (approved):
Everything looks good to me! Thanks for this!

corngood (approved):
Thanks!

gepbird:
Friendly ping @Mic92 :)

PR #328, opened 2025-03-12
feat: add snapshot support for sourcehut

antonmosich:
This adds support for following branches on sourcehut as well as tags. I'm new to this codebase and haven't really written python for a few years, but it worked for successfully for a package in nixpkgs.


SuperSandro2000 (commented):
Please add a test case

antonmosich:
I added a test now.
This was broken for a few days now due to [sourcehut having problems with LLM crawlers](https://status.sr.ht/issues/2025-03-17-git.sr.ht-llms/), which lead to the rss file not being accessible. This lead to me contacting sourcehut and asking whether they are fine with this approach, which they are.
We should add a User-Agent for the requests though, if only for sourcehut to be able to whitelist "us" in case they have to reimplement this tar pit again. I tried to do this, wanting to set the User-Agent to `nix-update/{version}`, but did not find a way to access the version of `nix-update`. How should I go about this?

Mic92:
@antonmosich we currently don't have a version in nix-update itself, however we can add this in our release script: https://github.com/Mic92/nix-update/blob/02beb5e1bc78bebc5996ca29b1333c4e77d5592d/bin/create-release.sh#L32 i.e. have a python file that exports `VERSION = "0.0.0"` and than is imported elsewhere.

antonmosich:
I struggled a bit with naming for this, as version is already used a lot all over the place.
I considered adding something like a `--print-version` flag, but didn't want to have too much going on in this PR

PR #330, opened 2025-03-15
Support --use-update-script with flake packages

matt1432:
1. like the title says, I made it possible to use `--use-update-script` with flake packages. However, I've never used that flag outside of my personal use so I'm really unsure if I implemented correctly
2. Added `--commit-message` flag to print the commit message to stdout
3. Added `--quiet` flag to suppress logs of subprocesses and stuff

### DIFF
--- a/nix_update/update.py
+++ b/nix_update/update.py
@@ -512,10 +512,8 @@ def update_version(
     return True
 
 
-def update(opts: Options) -> Package:
-    package = eval_attr(opts)
-
-    if package.has_update_script and opts.use_update_script:
+def run_update_script(package: Package, opts: Options) -> None:
+    if not opts.flake:
         run(
             [
                 "nix-shell",
@@ -523,10 +521,59 @@ def update(opts: Options) -> Package:
                 "--argstr",
                 "package",
                 opts.attribute,
+                "--argstr",
+                "skip-prompt",
+                "true",
                 *opts.update_script_args,
             ],
             stdout=None,
         )
+        return
+
+    update_script = run(
+        [
+            "nix",
+            "--extra-experimental-features",
+            "flakes nix-command",
+            "build",
+            "--print-out-paths",
+            "--impure",
+            "--expr",
+            f'with import <nixpkgs> {{}}; let pkg = {get_package(opts)}; in (pkgs.writeScript "updateScript" (builtins.toString (map builtins.toString (pkgs.lib.toList (pkg.updateScript.command or pkg.updateScript)))))',
+        ],
+    ).stdout.strip()
+
+    run(
+        [
+            "nix",
+            "develop",
+            "--impure",
+            "--expr",
+            f"with import <nixpkgs> {{}}; pkgs.mkShell {{inputsFrom = [{get_package(opts)}];}}",
+            "--command",
+            "bash",
+            "-c",
+            " ".join(
+                [
+                    "env",
+                    f"UPDATE_NIX_NAME={package.name}",
+                    f"UPDATE_NIX_PNAME={package.pname}",
+                    f"UPDATE_NIX_OLD_VERSION={package.old_version}",
+                    f"UPDATE_NIX_ATTR_PATH={package.attribute}",
+                    update_script,
+                ]
+            ),
+            *opts.update_script_args,
+        ],
+        cwd=opts.import_path,
+    )
+
+
+def update(opts: Options) -> Package:
+    package = eval_attr(opts)
+
+    if package.has_update_script and opts.use_update_script:
+        run_update_script(package, opts)
         new_package = eval_attr(opts)
         package.new_version = Version(
             new_package.old_version, rev=new_package.rev, tag=new_package.tag
--- a/tests/test_flake.py
+++ b/tests/test_flake.py
@@ -40,3 +40,40 @@ def test_main(helpers: conftest.Helpers) -> None:
         ).stdout.strip()
         print(diff)
         assert "https://diff.rs/fd-find/8.0.0/10.2.0" in diff
+
+
+def test_update_script(helpers: conftest.Helpers) -> None:
+    with helpers.testpkgs(init_git=True) as path:
+        main(
+            [
+                "--file",
+                str(path),
+                "--flake",
+                "--commit",
+                "--test",
+                "--use-update-script",
+                "crate",
+            ]
+        )
+        version = subprocess.run(
+            [
+                "nix",
+                "eval",
+                "--raw",
+                "--extra-experimental-features",
+                "flakes nix-command",
+                f"{path}#crate.version",
+            ],
+            check=True,
+            text=True,
+            stdout=subprocess.PIPE,
+        ).stdout.strip()
+        assert tuple(map(int, version.split("."))) >= (8, 5, 2)
+        commit = subprocess.run(
+            ["git", "-C", path, "log", "-1"],
+            text=True,
+            stdout=subprocess.PIPE,
+            check=True,
+        ).stdout.strip()
+        print(commit)
+        assert f"crate: 8.0.0 -> {version}" in commit
--- a/tests/testpkgs/crate.nix
+++ b/tests/testpkgs/crate.nix
@@ -2,6 +2,7 @@
   rustPlatform,
   fetchCrate,
   hello,
+  nix-update-script,
 }:
 rustPlatform.buildRustPackage rec {
   pname = "fd-find";
@@ -17,5 +18,11 @@ rustPlatform.buildRustPackage rec {
   passthru.tests = {
     foo = hello;
     bar = hello;
+  };
+  passthru.updateScript = nix-update-script {
+    attrPath = "crate";
+    extraArgs = [
+      "--flake"
+    ];
   };
 }
--- a/nix_update/eval.py
+++ b/nix_update/eval.py
@@ -40,6 +40,7 @@ class Package:
     attribute: str
     import_path: InitVar[str]
     name: str
+    pname: str
     old_version: str
     filename: str
     line: int
@@ -144,9 +145,7 @@ def eval_expression(
           sanitizePosition = x: x;
         """
 
-    has_update_script = (
-        "false" if flake else "pkg.passthru.updateScript or null != null"
-    )
+    has_update_script = "pkg.passthru.updateScript or null != null"
 
     return f"""
 let
@@ -170,6 +169,7 @@ let
     sanitizePosition (positionFromMeta pkg);
 in {{
   name = pkg.name;
+  pname = pkg.pname;
   old_version = pkg.version or (builtins.parseDrvName pkg.name).version;
   inherit raw_version_position;
   filename = position.file;
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -21,7 +21,8 @@ class Helpers:
     @staticmethod
     @contextmanager
     def testpkgs(init_git: bool = False) -> Iterator[Path]:
-        with tempfile.TemporaryDirectory() as tmpdirname:
+        with tempfile.TemporaryDirectory() as _tmpdirname:
+            tmpdirname = Path(_tmpdirname)
             shutil.copytree(
                 Helpers.root().joinpath("testpkgs"), tmpdirname, dirs_exist_ok=True
             )


nix_update/update.py
```diff
@@ -479,24 +494,38 @@ def update_version(
...
-                "package",
-                opts.attribute,
-                *opts.update_script_args,
-            ],
-            stdout=None,
-        )
+        if opts.flake:
+            run(
+                [
+                    "nix",
```
Mic92, line 507:
```suggestion
                    "nix",
                    "--extra-experimental-features", 
                    "flakes nix-command",
```

nix_update/__init__.py
```diff
@@ -53,6 +56,11 @@ def parse_args(args: list[str]) -> Options:
         "--url",
         help="URL to the repository to check for a release instead of using the URL in the src attribute of the package",
     )
+    parser.add_argument(
+        "--commit-message",
```
Mic92, line 60:
```suggestion
        "--print-commit-message",
```

Mic92:
Run `nix fmt` locally once to fix the CI error.

matt1432:
There you go

nix_update/update.py
```diff
@@ -479,24 +494,40 @@ def update_version(
...
+                    "--extra-experimental-features",
+                    "flakes nix-command",
+                    "run",
```
Mic92, line 510: That doesn't quite work. updateScript is defined as a list of arguments.

Here is one way to get this data:

```
/var/folders/_r/8fwc6pl57pq21rwsr9wz2b7h0000gn/T/tmp9n1gookj main
% nix --extra-experimental-features 'flakes nix-command' build --print-out-paths --impure --expr 'with import <nixpkgs> {}; pkgs.writeText "foo" (builtins.toString (builtins.getFlake "/private/var/folders/_r/8fwc6pl57pq21rwsr9wz2b7h0000gn/T/tmp9n1gookj#").packages.aarch64-darwin.crate.updateScript)'

warning: not writing modified lock file of flake 'path:/private/var/folders/_r/8fwc6pl57pq21rwsr9wz2b7h0000gn/T/tmp9n1gookj':
• Added input 'nixpkgs':
    'path:/nix/store/mfl5y74ksljrpwvprsxj21r9mzczma9z-source?lastModified=0&narHash=sha256-3wYmM6m1B29t80pGa/cNfIMRGM4K2AQQ51UNuUYf7j0%3D' (1970-01-01)
/nix/store/fvl66xg0wdcdx51kjml6mlf6gknjkw9s-foo

/var/folders/_r/8fwc6pl57pq21rwsr9wz2b7h0000gn/T/tmp9n1gookj main
% command cat /nix/store/fvl66xg0wdcdx51kjml6mlf6gknjkw9s-foo
/nix/store/ibn4hcgnmbcqa5hxd05q6xi8bf9ndmaw-nix-update-1.10.0/bin/nix-update
```

It would be also cool, if you could get rid of the `--impure` flag by passing `narHash` argument to `builtins.getFlake`. See this code here: https://git.clan.lol/clan/clan-core/src/commit/fcbdae9d0982eaa10f907e5b0f576805dfe489de/pkgs/clan-cli/clan_cli/flake.py#L456

And this code for computing the nar hash:

https://git.clan.lol/clan/clan-core/src/commit/fcbdae9d0982eaa10f907e5b0f576805dfe489de/pkgs/clan-cli/clan_cli/flake.py#L404

matt1432: > updateScript is defined as a list of arguments
where are those arguments called? I really don't understand how it works in nixpkgs tbh. Can you give me an example / what it looks like when it's working?

Mic92: Looks like it can be either a list or an attrset that contains the name "command"? https://github.com/NixOS/nixpkgs/blob/7344a3b78128f7b1765dba89060b015fb75431a7/maintainers/scripts/update.nix#L245

Mic92: And than it gets run here: https://github.com/NixOS/nixpkgs/blob/7344a3b78128f7b1765dba89060b015fb75431a7/maintainers/scripts/update.py#L198

matt1432: Ah so it depends on stuff from nixpkgs, I though it was all handled by nix-update and I just couldn't find the code responsible for running the script.

So is it out of scope to run the updateScript from nix-update itself?

We could simply do this without the environment variables and stuff with something like this:

```python
            update_script = run(
                [
                    "nix",
                    "--extra-experimental-features",
                    "flakes nix-command",
                    "build",
                    "--print-out-paths",
                    "--impure",
                    "--expr",
                    f'with import <nixpkgs> {{}}; let pkg = {get_package(opts)}; in (pkgs.writeScript "updateScript" (builtins.toString (map builtins.toString (pkgs.lib.toList (pkg.updateScript.command or pkg.updateScript)))))',
                ],
                quiet=opts.quiet,
            ).stdout.strip()

            run(
                [
                    "nix",
                    "develop",
                    "--command",
                    "bash",
                    "-c",
                    update_script,
                    *opts.update_script_args,
                ],
                quiet=opts.quiet,
            )
```

Mic92: It looks feasible to re-implement the environment that nixpkgs uses in nix-update. I don't expect it change that much in backwards-incompatible way.

matt1432: @Mic92 I figured out a way to use the `update.nix` file from nixpkgs directly and it worked on my personal tests. However, the test you added still fails and I don't really understand why. Any ideas?

tests/test_flake.py
```diff
@@ -40,3 +40,51 @@ def test_main(helpers: conftest.Helpers) -> None:
         ).stdout.strip()
         print(diff)
         assert "https://diff.rs/fd-find/8.0.0/10.2.0" in diff
+
+
+def test_update_script(helpers: conftest.Helpers) -> None:
+    with helpers.testpkgs(init_git=True) as path:
+        breakpoint()
```
Mic92, line 47:
```suggestion
```

nix_update/update.py
Mic92: Removed the breakpoint.

Mic92:
Can we skip prompt here?

  File "/nix/store/qli6b8pdydp2b9k7240qj9nb7vsn7csa-update.py", line 249, in <module>
    main(args.max_workers, args.keep_going, args.commit, args.packages, args.skip_prompt)
  File "/nix/store/qli6b8pdydp2b9k7240qj9nb7vsn7csa-update.py", line 223, in main
    confirm = '' if skip_prompt else input('Press Enter key to continue...')

Mic92:
New error:

```
input = None, capture_output = False, timeout = None, check = True
popenargs = (['nix-shell', '/nix/store/4b5yx89qafrc2izvv98fwnc87lwgl0mc-pbw8syks3fj41c7ha4vxmw7fjb0qqi93-source/maintainers/scripts/update.nix', '--impure', '--arg', 'skip-prompt', 'true', ...],)
kwargs = {'cwd': None, 'env': {'ALSOFT_DRIVERS': 'pipewire', 'ALTERNATE_EDITOR': 'nvim', 'AR': 'ar', 'AS': 'as', ...}, 'stderr': None, 'stdout': None, ...}
process = <Popen: returncode: 1 args: ['nix-shell', '/nix/store/4b5yx89qafrc2izvv98fwn...>, stdout = None
stderr = None, retcode = 1

    def run(*popenargs,
            input=None, capture_output=False, timeout=None, check=False, **kwargs):
        """Run command with arguments and return a CompletedProcess instance.

        The returned instance will have attributes args, returncode, stdout and
        stderr. By default, stdout and stderr are not captured, and those attributes
        will be None. Pass stdout=PIPE and/or stderr=PIPE in order to capture them,
        or pass capture_output=True to capture both.

        If check is True and the exit code was non-zero, it raises a
        CalledProcessError. The CalledProcessError object will have the return code
        in the returncode attribute, and output & stderr attributes if those streams
        were captured.

        If timeout is given, and the process takes too long, a TimeoutExpired
        exception will be raised.

        There is an optional argument "input", allowing you to
        pass bytes or a string to the subprocess's stdin.  If you use this argument
        you may not also use the Popen constructor's "stdin" argument, as
        it will be used internally.

        By default, all communication is in bytes, and therefore any "input" should
        be bytes, and the stdout and stderr will be bytes. If in text mode, any
        "input" should be a string, and stdout and stderr will be strings decoded
        according to locale encoding, or by "encoding" if set. Text mode is
        triggered by setting any of text, encoding, errors or universal_newlines.

        The other arguments are the same as for the Popen constructor.
        """
        if input is not None:
            if kwargs.get('stdin') is not None:
                raise ValueError('stdin and input arguments may not both be used.')
            kwargs['stdin'] = PIPE

        if capture_output:
            if kwargs.get('stdout') is not None or kwargs.get('stderr') is not None:
                raise ValueError('stdout and stderr arguments may not be used '
                                 'with capture_output.')
            kwargs['stdout'] = PIPE
            kwargs['stderr'] = PIPE

        with Popen(*popenargs, **kwargs) as process:
            try:
                stdout, stderr = process.communicate(input, timeout=timeout)
            except TimeoutExpired as exc:
                process.kill()
                if _mswindows:
                    # Windows accumulates the output in a single blocking
                    # read() call run on child threads, with the timeout
                    # being done in a join() on those threads.  communicate()
                    # _after_ kill() is required to collect that and add it
                    # to the exception.
                    exc.stdout, exc.stderr = process.communicate()
                else:
                    # POSIX _communicate already populated the output so
                    # far into the TimeoutExpired exception.
                    process.wait()
                raise
            except:  # Including KeyboardInterrupt, communicate handled that.
                process.kill()
                # We don't call process.wait() as .__exit__ does that for us.
                raise
            retcode = process.poll()
            if check and retcode:
>               raise CalledProcessError(retcode, process.args,
                                         output=stdout, stderr=stderr)
E               subprocess.CalledProcessError: Command '['nix-shell', '/nix/store/4b5yx89qafrc2izvv98fwnc87lwgl0mc-pbw8syks3fj41c7ha4vxmw7fjb0qqi93-source/maintainers/scripts/update.nix', '--impure', '--arg', 'skip-prompt', 'true', '--argstr', 'package', 'crate', '--arg', 'include-overlays', '[(final: prev: {crate = (let flake = builtins.getFlake "/private/var/folders/_r/8fwc6pl57pq21rwsr9wz2b7h0000gn/T/tmpxt_wu_28"; in flake.packages.${builtins.currentSystem}."crate" or flake."crate");})]']' returned non-zero exit status 1.
```

Mic92:
That one is hopefully easier to fix.
You can get a shell inside the test by adding the `breakpoint()` function
and run the test with `pytest -s tests/test_flake.py`

matt1432:
Everything looks good now, but the commit message doesn't show the changelogs for some reason

Mic92:
Might be fine. When using update scripts, we are missing informations.

tests/testpkgs/crate.nix
```diff
@@ -18,4 +19,12 @@ rustPlatform.buildRustPackage rec {
     foo = hello;
     bar = hello;
   };
+  passthru.updateScript = nix-update-script {
+    attrPath = "crate";
+    extraArgs = [
+      "--flake"
+      "--file"
+      "REPLACE_THIS"
```
Mic92, line 27: Mhm. This makes it pass the test, but what does this mean for users? Do they know need to replace this manually to match their local filesystem?

matt1432: Well this was necessary for the test because the location for the file is generated by the test. For users, they will know where the file to change will be

matt1432: @Mic92 is this okay or do you want to find another solution?

Mic92: It sounds a bit hacky. But I am also not sure how to use the current interface in a better way.

matt1432: How would it be incompatible?

Mic92: I don't think we can use maintainers/scripts/update.nix directly because the builtins.getFlake hack will turn your flake into a store path that cannot be modified directly. However at the end of the day this interface doesn't do that much magic and it should be possible to run those scripts directly just exporting the right environment variables.

matt1432: Ohhhh I understand now yeah. I'll work on implementing that

matt1432: @Mic92 I've implemented it. Am I missing anything for the environment?

matt1432:
@Mic92  Is anything else needed for this PR to get merged?

Mic92:
No. It's on me to review and test it.

PR #334, opened 2025-03-29
Improve diff url with github

gepbird:
The diff link in the generated commit message was broken in some cases, most notably containing `None`, missing/adding extra `v` as a version prefix.

Fixes https://github.com/NixOS/nixpkgs/issues/394072
- when there was a github link, previously it expected `fetchFromGitHub`, so it tried to read `src.rev` or  `src.tag` which doesn't exist when using plain `fetchurl`
- when the user specifies a fixed version, eg. `--version 11.41` (without a version prefix), it expected that the the tag/rev is called exactly that, but it can be `v11.41` or with other prefix

The solution is rather hacky and it's limited to github, there are multiple factors that make it hard to implement correctly:
- unclear division between user provided version, version got from eval, version got from fetching
- inconsistent naming: `package`/`version` are used interchangably, `new`/`old`/no-prefix is inconsistent, `version_number` containing more than numbers
- heuristics like [stripping "v" version prefix](https://github.com/gepbird/nix-update/blob/2eaf862d4eb4b119362379ce9c23e1c5478b4113/nix_update/update.py#L31-L32)...

In a few weeks/months when I'll have more time, I'd be happy to do a bigger refactor in this area, is it okay with you @Mic92?

### DIFF
--- a/nix_update/update.py
+++ b/nix_update/update.py
@@ -417,7 +417,11 @@ def update_nuget_deps(opts: Options) -> None:
 
 
 def update_version(
-    package: Package, version: str, preference: VersionPreference, version_regex: str
+    opts: Options,
+    package: Package,
+    version: str,
+    preference: VersionPreference,
+    version_regex: str,
 ) -> bool:
     if preference == VersionPreference.FIXED:
         new_version = Version(version)
@@ -457,6 +461,9 @@ def update_version(
             package.old_version = recovered_version
             return False
 
+    if not replace_version(package):
+        return False
+
     if package.parsed_url:
         if package.parsed_url.netloc == "crates.io":
             parts = package.parsed_url.path.split("/")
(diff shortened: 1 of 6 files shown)


nix_update/update.py
```diff
@@ -469,7 +476,29 @@ def update_version(
...
+                if match is not None:
+                    old_rev_tag = match.group(1)
+
+            new_rev_tag = new_version.rev or new_version.tag
+            if new_rev_tag is None:
+                # happens with fixed version preference (and possibly more situtations?)
+                new_package = eval_attr(opts)
+                new_rev_tag = new_package.rev or new_package.tag
+
+                print(new_package.parsed_url)
```
Mic92, line 493: Is this for debugging?

```suggestion
```

gepbird: Yes, I forgot about that one

nix_update/update.py
```diff
@@ -469,7 +476,28 @@ def update_version(
...
+                new_rev_tag = new_package.rev or new_package.tag
+
+                if new_rev_tag is None and new_package.parsed_url is not None:
+                    # happens when using fetchurl with a github link rather than using fetchFromGitHub
+                    regex = re.compile(".*/releases/download/(.*)/.*")
+                    match = regex.match(new_package.parsed_url.path)
+                    if match is not None:
+                        new_rev_tag = match.group(1)
+
+            package.diff_url = f"https://github.com/{owner}/{repo.removesuffix('.git')}/compare/{old_rev_tag}...{new_rev_tag}"
```
Mic92, line 500: We probably shouldn't set diff_url if either the old or new rev tag is None:

```suggestion
            if old_rev_tag is not None and new_rev_tag is not None:
                package.diff_url = f"https://github.com/{owner}/{repo.removesuffix('.git')}/compare/{old_rev_tag}...{new_rev_tag}"
```

PR #349, opened 2025-04-15
enhance matching of different outputHash format texts

h0nIg:


nix_update/update.py
```diff
@@ -127,11 +127,13 @@ def nix_prefetch(opts: Options, attr: str) -> str:
...
         stderr = res.stderr.strip()
+        # got:    xxx
+        # expected 'xxx' but got 'xxx'
+        regex = re.compile(r".*got:?\s*'?([^']*)('|$)")
         got = ""
         for line in stderr.split("\n"):
-            line = line.strip()
-            if line.startswith("got:"):
-                got = line.split("got:")[1].strip()
+            if match := regex.fullmatch(line):
```
Mic92, line 135: Can you show what cases we didn't detect before and now do?

Mic92:
```
FAILED tests/test_pnpm.py::test_update - nix_update.errors.UpdateError: failed to retrieve hash when trying to update pnpm.pnpmDeps
```

Mic92:
Looks like the CI error is unrelated: https://github.com/Mic92/nix-update/actions/runs/14485858932/job/40631202254?pr=349#step:5:1685

h0nIg:
@Mic92 fetchFromGithub does not support netrc file usage, as is does not run inside nix - fetchtree and friends does . The following solves the issue with private github installations and an ugly UI login forward which is part of GHES:

```
  fetchFromInternalGitHub = {
    githubBase,
    owner,
    repo,
    hash,
    tag ? null,
    rev ? null
  }: stdenv.mkDerivation(finalAttrs: rec {
       # fetchFromGitHub derivation name is always "source"
       name = "source";

       revWithTag = if tag != null then "refs/tags/${tag}" else rev;
       url = "https://${githubBase}/${owner}/${repo}/tarball/${revWithTag}";
       src =
         builtins.fetchTree {
           type = "tarball";
           url = url;

           # use hash of final derivation attributes, once it will get overwritten on the outer derivation
           narHash = finalAttrs.outputHash;
         };

       dontUnpack = true;
       buildPhase = ''
         cp -a ${src}/ $out/
       '';

       # fixed output derivation
       outputHashAlgo = "sha256";
       outputHashMode = "recursive";
       outputHash = hash;
     });
```

Mic92:
Ok. Can you add a test for builtins.fetchTree? i.e. you can copy github-tag.nix and just replace the fetcher in it.

h0nIg:
@Mic92 done

PR #352, opened 2025-04-16
Add quiet flag + print-commit-message flag

Mic92:
Part of https://github.com/Mic92/nix-update/pull/330


nix_update/__init__.py
```diff
@@ -53,6 +56,11 @@ def parse_args(args: list[str]) -> Options:
         "--url",
         help="URL to the repository to check for a release instead of using the URL in the src attribute of the package",
     )
+    parser.add_argument(
+        "--print-commit-message",
```
Mic92, line 60: @matt1432 only change I added is that --print-commit-message now implies --quiet

PR #363, opened 2025-04-30
disable treefmt on riscv64

Mic92:


treefmt.nix
```diff
@@ -6,21 +6,23 @@
     { pkgs, ... }:
     {
       treefmt = {
+        flakeCheck = pkgs.hostPlatform.system != "riscv64-linux";
```
SuperSandro2000, line 9: Why not use availableOn pattern?

Mic92: I think that broke at some point and currently it's a dependency deep in the dependency tree of treefmt, which renders riscv64-linux unusable until the next staging cycle.

PR #382, opened 2025-05-30
Add support for updating pubspec lock

ghost:
The `pubspecLock` is required by both `buildDartApplication` and `buildFlutterApplication`. When Git dependencies are involved, it's also necessary to define `gitHashes`. While this approach may not be universally reliable, it works for some of existing packages in nixpkgs.


nix_update/update.py
```diff
@@ -270,6 +270,104 @@ def update_cargo_lock(
...
+            res = run(
+                [
+                    "nix",
+                    "build",
+                    "--out-link",
+                    f"{tempdir}/result",
+                    "--impure",
+                    "--print-out-paths",
+                    "--expr",
+                    f"\nwith import <nixpkgs> {{}}; \n{get_package(opts)}.overrideAttrs (old: {{\n  nativeBuildInputs = (old.nativeBuildInputs or []) ++ [ pkgs.yq ];\n  postUnpack = ''\n    yq .  \"$sourceRoot/pubspec.lock\" > $out\n    exit\n  '';\n  outputs = [ \"out\" ];\n     separateDebugInfo = false;\n}})\n",
```
Mic92, line 323: Make this a multiline string that uses textwrap.dedent: https://docs.python.org/3/library/textwrap.html
This will be more readable.

nix_update/update.py
```diff
@@ -270,6 +270,104 @@ def update_cargo_lock(
...
+                encoding="utf-8",
+            ) as f:
+                run(
+                    [
+                        "nix",
+                        "run",
+                        "-L",
+                        "--impure",
+                        *opts.extra_flags,
+                        "github:NixOS/nixpkgs/nixos-unstable#yq",
```
Mic92, line 300: why do we do this here?

Mic92: I am bit worried about breaking stuff here. The yaml standard is not very well defined and very complex.

nix_update/update.py
```diff
@@ -270,6 +270,104 @@ def update_cargo_lock(
...
+                if pkg["source"] == "git":
+                    git_deps[pkg["description"]["resolved-ref"]] = (
+                        f"{name}",
+                        pkg["description"]["url"],
+                    )
+
+            hashes = dict(ThreadPoolExecutor().map(git_prefetch, git_deps.items()))
+
+    if hashes:
+        start_pattern = re.compile(r"^(?P<indent>\s*)gitHashes(?:\.[\w-]+)*\s*=")
```
Mic92, line 348: are we parsing yaml here with regex?

PR #401, opened 2025-07-03
support multiple capture groups in version-regex

wolfgangwalther:
This supports multiple capture groups in the `--version-regex` option. The groups will be concatenated to a single version delimited by `.`. This allows extracting versions from versioning schemes such as PostgreSQL's, which use `REL_17_0` tags etc., i.e. they use a different separator.


mweinelt:
Would appreciate a release to automate the updating of sqlalchemy, which uses `rel_\d+_\d+_\d+` tags.

SuperSandro2000:
If the last release wasn't that long ago, I would have just done it but that made unsure. @wolfgangwalther do you want to do it? 😓

mweinelt:
@Mic92 ping, as requested.

Issue #322, opened 2025-01-21
Default to version `0` when the version-regex does not match


mweinelt:
I want to implement automatic updates for [servo](https://github.com/servo/servo), which up until today has had no proper release.

This implies that the version should be `0-unstable-YYYY-MM-DD`, which is already supported by using `--version branch=main`.

There are however a number of tags on the repository, that do not reflect the actual package version, which complicates the matter.

 https://github.com/servo/servo/tags

```pytb
$ nix-update --build --commit servo --version branch=main\
[...]
Update 0-unstable-2025-01-14 -> selectors-v0.22.0-unstable-2025-01-21 in /home/hexa/git/nixos/master/pkgs/by-name/se/servo/package.nix
```

Here it picked one of the unrelated tags as the latest prior version for the unstable commit. Not great, but I can restrict the pattern used for valid versions:

```pytb
$ nix-update --build --commit servo --version-regex "v(.*)" --version branch=main
[...]
fetch https://github.com/servo/servo/commits/main.atom
fetch https://github.com/servo/servo/releases.atom
Traceback (most recent call last):
  File "/nix/store/7cy1pwj5acc4q32jwwql81h3c590r8ag-nix-update-1.9.0/bin/.nix-update-wrapped", line 9, in <module>
    sys.exit(main())
             ^^^^^^
  File "/nix/store/7cy1pwj5acc4q32jwwql81h3c590r8ag-nix-update-1.9.0/lib/python3.12/site-packages/nix_update/__init__.py", line 319, in main
    package = update(options)
              ^^^^^^^^^^^^^^^
  File "/nix/store/7cy1pwj5acc4q32jwwql81h3c590r8ag-nix-update-1.9.0/lib/python3.12/site-packages/nix_update/update.py", line 508, in update
    update_hash = update_version(
                  ^^^^^^^^^^^^^^^
  File "/nix/store/7cy1pwj5acc4q32jwwql81h3c590r8ag-nix-update-1.9.0/lib/python3.12/site-packages/nix_update/update.py", line 439, in update_version
    new_version = fetch_latest_version(
                  ^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/7cy1pwj5acc4q32jwwql81h3c590r8ag-nix-update-1.9.0/lib/python3.12/site-packages/nix_update/version/__init__.py", line 128, in fetch_latest_version
    raise VersionError(
nix_update.errors.VersionError: Not version matched the regex. The following versions were found:
[...]

PR #430, opened 2025-08-08
feat(github.py): use all tags for unstable updates

bryango:
Closes:
- #322

**Note:** the implementation feels a bit hacky, but it could be very useful in practice so here it is.

This patch hacks around the tag selection issue for unstable updates by simply appending `-unstable-xxxx-xx-xx` to all tags as potential candidates, which can then be filtered by `--version-regex`. Version 0 is included as a final fallback such that:

- To hardcore a 0-unstable-xxxx-xx-xx version one can simply do `--version-regex=(0-unstable-.*)`
- Generic version regex such as `--version-regex=jq-(.*)` also works and would lead to a canonical `x.x.x-unstable-xxxx-xx-xx` version string.

### DIFF
--- a/nix_update/version/github.py
+++ b/nix_update/version/github.py
@@ -152,7 +152,7 @@ def fetch_github_snapshots(
         else fetch_github_versions_from_feed
     )
     versions = version_fetcher(url, owner, repo)
-    latest_version = versions[0].number if versions else "0"
+    versions = [ version.number for version in versions ] + ["0"]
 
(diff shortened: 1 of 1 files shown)


bryango:
I believe you also have to specify `--version=branch`, otherwise nix-update will only look at the stable tags. I think this is documented in README.md.

**Update:** ah I see that you edited your post and figured it out as well haha. Nice.

PR #436, opened 2025-08-29
`--no-src`: init

9999years:
Add a `--no-src` option which lets you skip updating the `src` attribute. I have a Rust package with (effectively) `src = ./.;`, so when the version is updated (which edits the `Cargo.lock`, the `cargoHash` needs to be updated, but there is no `src` to update.


SuperSandro2000 (approved):
LGTM

PR #439, opened 2025-09-05
fix: Properly skip GitHub fetcher

jetjinser:
(In my understanding) The GitHub fetcher doesn't check `netloc`, causing the gitea / codeberg URL path to be captured by the regex and enter the GitHub fetcher processing path.

This PR attempts to fix this issue by checking `netloc`.

**reproduction method**
in the nixpkgs repo
```sh
nix run github:Mic92/nix-update -- bibiman
```


SuperSandro2000:
fyi @h0nIg

PR #440, opened 2025-09-07
fix: escape shell args when using --flake and -use-update-script

gepbird:
Some update scripts worked with `--use-update-script`, but not with `--use-update-script --flake`, because it used a different script rendering logic where arguments wasn't escaped in the generated shell script.

An example of a generated incorrect update script: `/nix/store/6fxc7xv4czpqangbdwi14pqhw6jm9r5v-nix-update-1.12.1/bin/nix-update --version-regex ^(\\d{4}-\\d{2}-\\d{2})$` it results in this error: https://github.com/gepbird/nur-packages/actions/runs/17362727999/job/49284958152#step:4:812

After this PR shell arguments are escaped: `/nix/store/6fxc7xv4czpqangbdwi14pqhw6jm9r5v-nix-update-1.12.1/bin/nix-update --version-regex '^(\d{4}-\d{2}-\d{2})$'`.

## Summary by CodeRabbit

* **New Features**
  * Better flake integration: support for packages exposing configurable update scripts and date-formatted versions.
* **Bug Fixes**
  * Improved quoting/escaping of generated update scripts to handle special characters more safely.
* **Tests**
  * Added end-to-end test validating --use-update-script with flake workflows; test invocation now includes an explicit --version flag.
* **Chores**
  * Added test package fixtures to exercise the flake/update-script flow.

### DIFF
--- a/tests/test_flake.py
+++ b/tests/test_flake.py
@@ -6,7 +6,18 @@ from tests import conftest
 
 def test_main(helpers: conftest.Helpers) -> None:
     with helpers.testpkgs(init_git=True) as path:
-        main(["--file", str(path), "--flake", "--commit", "--test", "crate"])
+        main(
+            [
+                "--file",
+                str(path),
+                "--flake",
+                "--commit",
+                "--test",
+                "--version",
+                "10.2.0",
+                "crate",
+            ],
+        )
         version = subprocess.run(
             [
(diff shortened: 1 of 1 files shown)


SuperSandro2000 (approved):
LGTM

PR #469, opened 2025-09-26
mark canary releases as unstable

mdaniels5757:
I ran across this in fermyon-spin when working on 443625.

I also sorted the regex alphabetically.

## Summary by CodeRabbit

* **New Features**
  * Version stability detection now recognizes “canary” as an unstable marker, alongside alpha, beta, nightly, prerelease, preview, rc, and mN patterns. Versions containing these tokens (case-insensitive) are treated as unstable when no prerelease component is present, improving the accuracy of stability checks and any automations or workflows that rely on them.

### DIFF
--- a/nix_update/version/__init__.py
+++ b/nix_update/version/__init__.py
@@ -106,7 +106,7 @@ def extract_version(version: Version, version_regex: str) -> Version | None:
 def is_unstable(version: Version, extracted: str) -> bool:
     if version.prerelease is not None:
         return version.prerelease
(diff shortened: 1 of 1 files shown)


mdaniels5757:
I don't think the test failure is caused by me? It's caused by https://github.com/vrana/adminer/releases/download/v5.4.1/adminer-5.4.1.zip 404-ing, and I can't think of a way that this change could have caused it. That repo does not have any tag that isn't `v${version}`.

SuperSandro2000 (approved):

PR #501, opened 2025-11-17
fix fetchTarball output case

h0nIg:
we recently switched from fetchTree to fetchTarball for private github installations. unfortunately fetchTarball uses a different hash display schema:

```
       error: hash mismatch in file downloaded from 'https://xxx/api/v3/repos/xx/xx/tarball/refs/tags/v0xxx':
         specified: sha256:0000000000000000000000000000000000000000000000000000
         got:       sha256:00qz12iwzbh5bv3szvnqnq2a1c866v038z53i69jba74pwclhppg
```

### DIFF
--- a/nix_update/dependency_hashes.py
+++ b/nix_update/dependency_hashes.py
@@ -34,8 +34,8 @@ def extract_hash_from_nix_error(stderr: str) -> str | None:
 
     Handles various formats:
     - got:    xxx
-    - got:    sha256-xxxxx=
     - got:    sha256:xxxxx
+    - got:    sha256-xxxxx=
     - got:    blake3-xxxxx=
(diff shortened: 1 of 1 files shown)


nix_update/dependency_hashes.py
```diff
@@ -35,14 +35,15 @@ def extract_hash_from_nix_error(stderr: str) -> str | None:
     Handles various formats:
     - got:    xxx
     - got:    sha256-xxxxx=
+    - got:    sha256:xxxxx
```
SuperSandro2000:
```suggestion
    - got:    sha256:xxxxx
    - got:    sha256-xxxxx=
```
lets keep the sri and none sri hashes separate

PR #506, opened 2025-11-23
Pass desired output aarch as eval-system argument to nix

### LINKED ISSUE #479: Feature request: update system-dependent hashes
I maintain a package that uses a fixed-output derivation submodule which stores platform-specific hashes as an attribute set in `outputHash` (example: https://github.com/NixOS/nixpkgs/blob/029b4824cbfc2007f1c5834fc717d4ff3b1d6636/pkgs/by-name/op/opencode/package.nix#L106-L113).

Currently, to update a hash I have to run `nix-build <package>.<submodule> --system <platform>` and then replace the hashes manually. I saw that many other packages with system-dependent hashes provide a custom update script.

Wouldn't it be possible to detect when the output hash is an attribute set. If so, read the attribute keys (the systems) and update the hashes for all defined systems automatically.

Questions / notes:
- I looked through the code but didn’t find an obvious way to add this. Any pointers on how you’d implement it?
- Is this a niche problem, or something others encounter often?

Thanks for considering — happy to help prototype if you point me to the right place in the code.

Bert-Proesmans:
This change activates remote builders to build packages for other platforms instead of forcing the current host to be target platform-compatible (feature introduced in #130).

**It's unclear to me if this is also a solution to the original issue (https://github.com/Mic92/nix-update/issues/129) or a separate parameter on nix-update `eval-system` is required/desired.**

Fixes #479.

This change updates the following pseudo package recipe as intuitively expected;

```nix
{
  lib,
  stdenv,
  fetchurl,
  _experimental-update-script-combinators,
  nix-update-script,
}:
let
  version = "2.0.88";
  sourceMap = {
    aarch64-linux = fetchurl {
      url = "https://github.com/netbootxyz/netboot.xyz/releases/download/${version}/netboot.xyz-arm64.efi";
      hash = "sha256-AeW92FU65XVJKGPi+A/iz7Jvtb7wKIO3xG3Cx7v4kRg=";
    };
    x86_64-linux = fetchurl {
      url = "https://github.com/netbootxyz/netboot.xyz/releases/download/${version}/netboot.xyz.efi";
      hash = "sha256-ipbZJ0mPCuwzb/TDtXXUBTuWOcSsKGAJ1GEGIgB2G7E=";
    };
  };
in
stdenv.mkDerivation (finalAttrs: {
  pname = "netboot.xyz-efi";
  inherit version;

  src =
    sourceMap.${stdenv.hostPlatform.system}
      or (throw "Unsupported system: ${stdenv.hostPlatform.system}");

  passthru.updateScript = _experimental-update-script-combinators.sequence [
    (nix-update-script { extraArgs = ["--system" "x86_64-linux"]})
    (nix-update-script { extraArgs = ["--system" "aarch64-linux" "--version" "skip"]})
  ];
  # ---
})
```

into 

```nix
# ---
 sourceMap = {
    version = "2.0.89";
    aarch64-linux = fetchurl {
      url = "https://github.com/netbootxyz/netboot.xyz/releases/download/${version}/netboot.xyz-arm64.efi";
      hash = "sha256-N1s2l3+8zPX1eBKtKIVaKqr7DDUiUXlu/0lNR1EJmRs=";
    };
    x86_64-linux = fetchurl {
      url = "https://github.com/netbootxyz/netboot.xyz/releases/download/${version}/netboot.xyz.efi";
      hash = "sha256-rmMU/OTOh7zId+zwQCXHt2SJyBAtLNfCOZhDyrkBKjw=";
    };
  };
# ---
```

### DIFF
--- a/nix_update/__init__.py
+++ b/nix_update/__init__.py
@@ -167,7 +167,7 @@ def parse_args(args: list[str]) -> Options:
     a = parser.parse_args(args)
     extra_flags = ["--extra-experimental-features", "flakes nix-command"]
     if a.system:
-        extra_flags.extend(["--system", a.system])
+        extra_flags.extend(["--eval-system", a.system])
     for name, value in a.option:
         extra_flags.extend(["--option", name, value])
 


Bert-Proesmans:
Can you be more specific about use-case?

When I configure `boot.binfmt.emulatedSystems` with "aarch64-linux" and disable `nix.distributedBuilds`, the updater performs the expected hash update (matches real aarch64-linux build machine result) for aarch64-linux.
I deduce your example should "just work" with this PR change unless I have a blind spot.

PR #550, opened 2026-02-15
chore: Add Python with mypy, pytest, and ruff to devShell

lafrenierejm:
This adds the tools directly to `PATH`, allowing developers to invoke them directly without `nix develop -c`.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -40,11 +40,11 @@
               packages = [
                 (pkgs.python3.withPackages (
                   ps: with ps; [
-                    mypy
                     pytest
-                    ruff
                   ]
                 ))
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -37,7 +37,15 @@
 
             devShells.default = pkgs.mkShell {
               inputsFrom = [ config.packages.default ];
-
+              packages = [
+                (pkgs.python3.withPackages (
+                  ps: with ps; [
+                    mypy
+                    pytest
+                    ruff
```
Mic92: ruff and mypy comes from pkgs, not from pythonPackages.

PR #559, opened 2026-02-23
Sort fetched versions by semver instead of feed order

### LINKED ISSUE #492: nix-udate seems to prefer the latest published version by date and not by version number
# Description

Sometimes, some project releases a patch release for an old version after a minor release is made. nix-update seems to pick up the patch version instead of doing a version comparison sort.

Example: [postgis](https://github.com/postgis/postgis/tags) (3.5.4 published after 3.6), which then triggered the update bot: https://github.com/NixOS/nixpkgs/pull/453826 and https://github.com/NixOS/nixpkgs/pull/455937

# Possible solutions

It seems that there is no sort [here](https://github.com/Mic92/nix-update/blob/main/nix_update/version/__init__.py#L158). We should order according to semver.

vdemeester:
## Problem

`fetch_latest_version` returns the first version from the upstream feed that passes the regex filter. Since GitHub/GitLab feeds are ordered by **release date**, this causes downgrades when upstream publishes patch releases on older branches after newer versions exist (e.g. `v0.39.5` released after `v0.41.1`).

The `--version-regex` workaround only filters versions — it doesn't change the sort order.

Examples: NixOS/nixpkgs#492517, NixOS/nixpkgs#461118, NixOS/nixpkgs#453826

## Solution

Wires in the `version_compare` module from @Mic92's `version-compare` branch to sort candidate versions by semver (highest first) before selecting.

Fixes #492

### DIFF
--- a/nix_update/version/__init__.py
+++ b/nix_update/version/__init__.py
@@ -3,13 +3,14 @@ from __future__ import annotations
 import re
 from collections.abc import Callable
 from dataclasses import dataclass, field
-from functools import partial
+from functools import cmp_to_key, partial
 from inspect import signature
 from typing import Any, Protocol, cast
 from urllib.parse import ParseResult
 from urllib.request import build_opener, install_opener
 
 from nix_update.errors import VersionError
+from nix_update.version_compare import version_compare
 from nix_update.version_info import VERSION
 
 from .bitbucket import fetch_bitbucket_snapshots, fetch_bitbucket_versions
@@ -187,6 +188,14 @@ def fetch_latest_version(
         all_filtered.extend(filtered)
 
         if final:
+            # Sort by version number (highest first) to avoid picking
+            # older versions that were released more recently (e.g. patch
+            # releases on older branches).
+            final.sort(
+                key=cmp_to_key(
+                    lambda a, b: version_compare(b.number, a.number),
+                ),
+            )
             prefixed_version = find_prefixed_version(final, config)
             if prefixed_version is not None:
                 return prefixed_version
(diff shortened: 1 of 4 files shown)


kira-bruneau:
With this change it looks like if a project previously tagged versions with an extra dot like `v.1.0.0` instead of `v2.0.0`, the older version will be prioritized over the newer one (https://github.com/tgcowell/waves/tags is an example, where `v.2.2.0` will be prioritized over `v2.4.1`).

The root cause doesn't exactly seem to be with the version compare implementation added here, but how versions are parsed if you don't explicitly define a `--version-regex` option. Specifically `.2.2.0` & `2.4.1` (without the v) will get passed as strings to the `version_compare` function.

Maybe in this specific case we should treat `.2.2.0` as a version "lower" than `2.4.1`, which kind of makes sense, but it would be a hacky workaround, and wouldn't work if the format switched the other way (or with any other arbitrary tag inconsistencies).

In cases like this where the tag format changes throughout the history, I think the best solution would be to add an explicit `--version-regex` to handle both formats. So I'm not sure if this warrants a new issue, but I just wanted to leave this as a record in case anyone else runs into the same problem.

<hr>

_This is what I ended up using_:
https://github.com/kira-bruneau/nur-packages/blob/7b1935cd7455ce98647b6e9fd387e3fd39a83be1/pkgs/by-name/wa/home-assistant-waves-theme/package.nix#L25-L31

figsoda:
i think we should let `version_compare` handle this case by deprioritizing the non-numeric prefix, and compare the numeric part before the prefix, so `a-2.0.0 > b-1.0.0`, and `a-1.0.0 < b-1.0.0`

PR #580, opened 2026-04-09
http.fetch_json: support headers, gitlab: use GITLAB_TOKEN

Prince213:
This will allow updating private GitLab repositories.

We might further want a generic version of in `.version.github._dorequest` in `.version.http`.


Mic92:
Applied in https://github.com/Mic92/nix-update/pull/621

PR #594, opened 2026-05-06
fix: preserve prefixed GitHub release selection

### LINKED ISSUE #593: `version_compare` incorrectly ranks version-prefixed tags (e.g. `v4.10.7`) lower than bare numeric ones (e.g. `0.1.0`)
Since v1.15.0, `fetch_latest_version` sorts candidates with `version_compare` before selecting. This breaks when a repository has releases with mixed prefix conventions — for example `v4.10.7` (historical series) and newly-added `0.1.0` / `0.0.2` (new series without v prefix).

`version_compare.py`'s `VersionSegment.compare_to` treats any numeric first segment as unconditionally newer than any alphabetic first segment:

``` py
if self.is_numeric != other.is_numeric:
    return 1 if self.is_numeric else -1
```

This means `0.1.0` (first segment: numeric `0`) is judged newer than `v4.10.7` (first segment: alpha `v`), even though `4.10.7` >> `0.1.0`.

**Reproduction:** `zed-industries/vscode-langservers-extracted` — current package at `v4.10.7` (Nov 05 2025), releases feed now contains `0.1.0` (May 21 2025) and `0.0.2` (May 16 2025). nix-update selects `0.1.0` as latest, tries to fetch `v0.1.0.tar.gz`, gets 404, and fails.

bet4it:
Fixes #593.

## Summary
- preserve the highest matching prefixed release after version sorting
- avoid falling back to bare release tags when the current prefixed tag is already latest
- add regression coverage using the mixed-prefix reproducer from #593

## Testing
- pytest -o addopts='' -q tests/test_version_sorting.py
- pytest -o addopts='' -q tests/test_version_compare.py

### DIFF
--- a/tests/test_version_sorting.py
+++ b/tests/test_version_sorting.py
@@ -11,6 +11,7 @@ publishes v0.39.5 after v0.41.1).
 from __future__ import annotations
 
 import io
+import json
 import typing
 import unittest.mock
 import xml.etree.ElementTree as ET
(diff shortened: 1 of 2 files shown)


tests/test_version_sorting.py
```diff
@@ -136,3 +156,25 @@ def test_postgis_scenario() -> None:
             ),
         )
         assert version.number == "3.6.0"
+
+
+def test_mixed_prefixed_and_bare_github_releases_keep_prefixed_series() -> None:
```
figsoda, line 161: this pr does fix the reproduction case in #593, but doesn't fix the version_compare issue mentioned in the issue, which is fine, but this test will start passing once #600 is merged, and wouldn't really test what this pr does - letting version prefix work when no updates are available, not exactly a `version_sorting` issue

the change is simple enough, i think it's fine to leave out the tests

PR #600, opened 2026-05-16
version_compare: compare non-numeric prefix last

### LINKED ISSUE #593: `version_compare` incorrectly ranks version-prefixed tags (e.g. `v4.10.7`) lower than bare numeric ones (e.g. `0.1.0`)

### LINKED ISSUE #597: --version=branch algorithm changed/broken
see: https://github.com/r-ryantm/nixpkgs/commits/6afafa2c6e0b24e83a0b7cbb8815c171b99d2adf/pkgs/by-name/st/steam-devices-udev-rules/package.nix

https://github.com/NixOS/nixpkgs/pull/518868

This used to work just fine, but now the string extraction is not working like it used to.

figsoda:
fixes #597 

fixes #593 

also see https://github.com/Mic92/nix-update/pull/559#issuecomment-4450642704

### DIFF
--- a/nix_update/version_compare.py
+++ b/nix_update/version_compare.py
@@ -12,6 +12,10 @@ The implementation is inspired by RPM-style version comparison but is an
 independent implementation written specifically for nix-update.
 """
 
+import re
+
+DIGIT = re.compile(r"\d")
+
 
 def _cmp_str(a: str, b: str) -> int:
     """Return -1, 0, or 1 comparing two strings lexicographically."""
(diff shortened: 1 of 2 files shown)


figsoda:
@kira-bruneau this should handle the comparison case, didn't change the prefix stripping logic since i didn't want to accidentally break things, so `.1.2.3` will still show up with the `.` prefix if it ends up being the latest

kira-bruneau (approved):
Thanks @figsoda! This fixes the issue I was running into.

yuannan:
Do you know why the bot is still having this issue?

https://github.com/NixOS/nixpkgs/pull/524223#issuecomment-4623157767

figsoda:
@yuannan the package was updated using an older version of nix-update https://github.com/NixOS/nixpkgs/blob/c339eccfbba7a37cc13c9aee40f704094546cc93/pkgs/by-name/ni/nix-update/package.nix

PR #607, opened 2026-05-28
update_dependency_hashes: workaround goModules hash depending on other hashes

### LINKED ISSUE #302: sub-hashes can be wrong if they depend on each other
`nix_update/update.py`'s `update` updates hashes in a fixed order (`update_go_modules_hash`, `update_cargo_vendor_deps_hash`...)
but if multiple such hashes are required, it's possible that the earlier hash actually depends on a later one.

To give a concrete example, rmfakecloud as of https://github.com/NixOS/nixpkgs/pull/358324/commits/632b589b55eec8fcbad9045e6e51338962ea1bf4 contains a go module with pnpmDeps.

building rmfakecloud.goModules pulls in pnpmDeps so building it with a fake hash fails due to a bad hash in pnpmDeps; and the update script will take that pnpmDep's hash and enter it in the go's `vendorHash`

Just changing the way the hashes are evaluated fixes this issue:

```diff
diff --git a/nix_update/update.py b/nix_update/update.py
index 605ac2f6f3bb..9dd8eaf73939 100644
--- a/nix_update/update.py
+++ b/nix_update/update.py
@@ -469,6 +469,9 @@ def update(opts: Options) -> Package:
 
     # if no package.hash was provided we just update the other hashes unconditionally
     if update_hash or not package.hash:
+        if package.pnpm_deps:
+            update_pnpm_deps_hash(opts, package.filename, package.pnpm_deps)
+
         if package.go_modules:
             update_go_modules_hash(opts, package.filename, package.go_modules)
 
@@ -494,9 +497,6 @@ def update(opts: Options) -> Package:
         if package.npm_deps:
             update_npm_deps_hash(opts, package.filename, package.npm_deps)
 
-        if package.pnpm_deps:
-            update_pnpm_deps_hash(opts, package.filename, package.pnpm_deps)
-
         if package.yarn_deps:
             update_yarn_deps_hash(opts, package.filename, package.yarn_deps)
 
```

But that'll obviously break updating pnpmDeps hash if that hash depends on a go module (I don't think that's likely, but it's technically possible...)

So, would it make sense/be possible to interface something to specify the order in which we want these to run?
Better, could we detect they depend on each other and compute an appropriate order automatically?

(An alternative less efficient than controlling the order would be to try to run these multiple times (e.g. updating go, pnpm, go would ultimately fill in the correct go hash), but that's a bit too much of a kludge in my opinion so not suggesting that one)

btw, I don't think such package is common enough to warrant complexifying the code too much, mostly wondering.

Thanks!

martinetd:
Some packages like rmfakecloud cannot be reliably updated, this makes the autoupdate script work and create content similar to this PR done manually[1]

This has been discussed in #302 before but the goModules rework that was drafted at that point went stale, and we might as well work around the problem[2] in nix-update: basically, the goModules hash is based on the whole derivation, and not just the actual go dependencies being pulled, so we want to keep it late in execution order.

Link: https://github.com/NixOS/nixpkgs/pull/524309 [1]
Link: https://github.com/NixOS/nixpkgs/issues/358844 [2]
Fixes #302

---------------

Hi! I brought this up for discussion in #302 two years ago and basically we decided back then it's "not our bug", but given there's no fix in sight I think a workaround would make sense.

In theory it's possible this PR could break some other updates using goModules with something else that also has this cyclic dependency problem, but if there were such other modules then they'd never possibly settle on a valid hash (updating go would update the other's hash, and in turn updating that other's hash would update go's), so I'm pretty confident this shouldn't break anything.
It's just ugly and not mentally satisfying.

(cc @coolGi69 as he's spent time looking at it again recently - thank you!)

### DIFF
--- a/nix_update/dependency_hashes.py
+++ b/nix_update/dependency_hashes.py
@@ -170,10 +170,14 @@ def update_dependency_hashes(
     if not (update_hash or not package.hash) or opts.src_only:
         return
 
+    # In theory dependency hashes should only depend on the actual dependencies
+    # being fetched, but some derivation frameworks like goModules pull in the
+    # whole derivation in which case updating other dependencies end up
+    # modifying the hash. Ensure such modules come "later" in order below.
+    # https://github.com/NixOS/nixpkgs/issues/358844
+    # Dictionary iteration order is guaranteed since python 3.7
     hash_updaters: dict[str, Callable[[Options, str, Any], None]] = {
         "fod_subpackage": partial(update_hash_with_prefetch, None),
-        "go_modules": partial(update_hash_with_prefetch, "goModules"),
-        "go_modules_old": partial(update_hash_with_prefetch, "go-modules"),
         "cargo_deps": partial(update_hash_with_prefetch, "cargoDeps"),
         "cargo_vendor_deps": partial(
             update_hash_with_prefetch,
@@ -189,6 +193,8 @@ def update_dependency_hashes(
         "mix_deps": partial(update_hash_with_prefetch, "mixFodDeps"),
         "zig_deps": partial(update_hash_with_prefetch, "zigDeps"),
         "cargo_lock": update_cargo_lock,
+        "go_modules": partial(update_hash_with_prefetch, "goModules"),
+        "go_modules_old": partial(update_hash_with_prefetch, "go-modules"),
     }
 
     # Update all dependency hashes using registry


martinetd:
(test failure looks unrelated:
>        > curl: (22) The requested URL returned error: 403
>        > error: cannot download fd-find-10.2.0.tar.gz from any mirror
)

nix_update/dependency_hashes.py
```diff
@@ -153,10 +153,14 @@ def update_dependency_hashes(
     if not (update_hash or not package.hash) or opts.src_only:
         return
 
+    # In theory dependency hashes should only depend on the actual dependencies
+    # being fetched, but some derivation frameworks like goModules pull in the
+    # whole derivation in which case updating other dependencies end up
+    # modifying the hash. Ensure such modules come "later" in order below.
+    # https://github.com/NixOS/nixpkgs/issues/358844
+    # Dictionary iteration order is guaranteed since python 3.7
```
SuperSandro2000, line 178:
```suggestion
```

SuperSandro2000:
Does this only effect go?

nix_update/dependency_hashes.py
```diff
@@ -172,6 +176,8 @@ def update_dependency_hashes(
         "mix_deps": partial(update_hash_with_prefetch, "mixFodDeps"),
         "zig_deps": partial(update_hash_with_prefetch, "zigDeps"),
         "cargo_lock": update_cargo_lock,
```
SuperSandro2000, line 195: We should mention here in a comment that the go parts are special

nix_update/dependency_hashes.py
martinetd: This is far from obvious to me, but if you think it is then I guess we can remove the comment (this whole thing doesn't make sense if order is not guaranteed)

nix_update/dependency_hashes.py
martinetd: > Does this only effect go?

I honestly don't know -- I do know pnpm_deps doesn't have the problem otherwise the derivation wouldn't be buildable in the first place, but I haven't checked (and wouldn't know how to check in a reasonable time) other frameworks
(FWIW, I don't even know if go_modules_old had the problem, I just decided it'd make more sense to keep together...)

> We should mention here in a comment that the go parts are special

In that case let's remove the comment above, and describe the problem here explaining the issue and saying "anything below this line include the whole derivation in the hash and must come late in order to allow nested (anything above) to work" ?

PR #610, opened 2026-06-06
Handle static.crates.io domain

Defelo:
See https://github.com/NixOS/nixpkgs/pull/525067

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -22,11 +22,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1779536132,
-        "narHash": "sha256-q+fF42iv/geEbHfgSzy3tS0FF/EyD6XTZ98E6yxiBO8=",
+        "lastModified": 1780747962,
+        "narHash": "sha256-IX7G1dlKrOqPOImfbo7ADDfV5yU1+j+MRChI3TL4tAA=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "3d8f0f3f72a6cd4d93d0ad13203f2ea1cb7e1456",
+        "rev": "cbb5cf358f50aa6acc9efd6113b7bcfbc352cd73",
         "type": "github"
       },
       "original": {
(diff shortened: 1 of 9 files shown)


SuperSandro2000:
Please take a look at the falling tests

Defelo:
> Please take a look at the falling tests

I have made several attempts to fix the tests yesterday (see force push history), but every time I thought I fixed a test failure more tests started to fail for seemingly no reason. Most of these failures now seem unrelated to the changes I made. Right now the only idea I have left is to rerun the tests and hope they magically succeed, so I would appreciate some help here...

Defelo:
Nevermind, apparently I was just blind :see_no_evil: 

Tests are green now

SuperSandro2000 (commented):
How hard would it be to adapt/add a test to also test this?

Defelo:
Given that the tests just use `fetchCrate` from nixpkgs, it's just a matter of updating the nixpkgs flake input (after merging this PR, creating a new release, and updating nix-update in nixpkgs)

Defelo:
> @SuperSandro2000 pushed 1 commit.

@SuperSandro2000 that's what I tried before, but the tests actually use nix-update from nixpkgs (which obviously doesn't know about static.crates.io yet).

SuperSandro2000:
I do not understand why the last test fails 😅

Defelo:
> I do not understand why the last test fails 😅

@SuperSandro2000 This test runs nix-update with the `--use-update-script` flag on `testpkgs#crate`, so it actually just executes `testpkgs#crate.updateScript` which itself just uses nix-update again (via the `nix-update-script` function in nixpkgs), but not the nix-update we have here, but the nix-update defined in nixpkgs (i.e. an older nix-update which cannot handle the static.crates.io domain yet which `testpkgs#crate.src` uses now because you updated the flake input). Ideally the `nix-update-script` in `testpkgs` should refer to the local version of `nix-update` and not the one defined in `nixpkgs`, not sure why this hasn't been implemented this way though.

PR #626, opened 2026-06-11
rip: mergify/renovate

Mic92:
--template now requires an


06kellyjac:
"--template now requires an" what? I'm curious now :sweat_smile:

